Issue #36, opened 2023-07-11
Node should check the machine hash from blockchain


gligneul:
## 📚 Context

Currently, the node doesn't check whether the initialized Cartesi Machine has the same hash of the DApp contract deployed in the blockchain. The node assumes that the validator deployer will perform this check.

We have the premise that we should trust the validator. However, the validator might use an invalid machine due to a mistake, vulnerability, or bug. So, the node should double-check the hash to ensure it is using the correct Cartesi machine.

## ✔️ Solution

After initializing the Cartesi Machine, the node should check whether the local machine has the same hash as the deployed DApp in the blockchain. 
The service that should do this check is the advance runner, which is responsible for initializing the server manager. The advance-runner should connect to the blockchain and get the deployed hash to do that.

## 📈 Subtasks

- [x] Obtain the expected hash from the DApp contract.
- [x] Obtain the hash from the initialized Cartesi Machine.
- [x] Compare both hashes and raise an error if they are not the same.

PR #48, opened 2023-08-01
feat(advance-runner): get off-chain machine hash

torives:
This PR is the halfway implementation of #36. Currently, there's no check, we're just logging the hash obtained from the Cartesi Machine snapshot. A second PR will follow with the on-chain hash and the actual comparison.


gligneul (approved):
Looks great!

renan061 (approved):

PR #52, opened 2023-08-08
Support POST requests at the inspect-server

### LINKED ISSUE #27: Add POST inspect to HTTP API
## 📚 Context

Currently, the inspect HTTP API has only one endpoint that receives the payload in the request path. This is limiting because it is not possible to send binaries as a payload, and there is a size constraint to the size of the payload.

## ✔️ Solution

Add another endpoint to the inspect API that receives the payload in the body. This endpoint should be a POST because we shouldn't receive a body in an HTTP get request. We shouldn't remove the old endpoint; it might still be useful for those who encode the inspect payload as a path.

## 📈 Subtasks

- [x] Define POST inspect in the OpenAPI interfaces.
- [x] Add new endpoint to the inspect HTTP server.
- [x] Update tests
- [ ] Update documentation

marcelstanley:
- [x] Added support for POST requests
- [x] Replicated most of the existing tests to also cover POST requests
- [x] Updated API at companion PR cartesi/openapi-interfaces#2

Closes #27

### DIFF
--- a/offchain/inspect-server/src/server.rs
+++ b/offchain/inspect-server/src/server.rs
@@ -14,20 +14,29 @@ use crate::inspect::{
     CompletionStatus, InspectClient, InspectStateResponse, Report,
 };
 
+// (2^20 - 64) bytes, which is the length of the RX buffer minus metadata
+pub const CARTESI_MACHINE_RX_BUFFER_LIMIT: usize = 1_048_512;
+
 pub fn create(
     config: &InspectServerConfig,
     inspect_client: InspectClient,
 ) -> std::io::Result<Server> {
-    let inspect_path = config.inspect_path_prefix.clone() + "/{payload:.*}";
+    let inspect_path = config.inspect_path_prefix.clone();
+    let inspect_get_path = inspect_path.clone() + "/{payload:.*}";
     let server = HttpServer::new(move || {
         let cors = Cors::permissive();
         App::new()
             .app_data(web::Data::new(inspect_client.clone()))
+            .app_data(web::PayloadConfig::new(CARTESI_MACHINE_RX_BUFFER_LIMIT))
             .wrap(middleware::Logger::default())
             .wrap(cors)
             .service(
+                web::resource(inspect_get_path.clone())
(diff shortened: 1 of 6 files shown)


offchain/inspect-server/src/server.rs
```diff
@@ -18,7 +18,8 @@ pub fn create(
     config: &InspectServerConfig,
     inspect_client: InspectClient,
 ) -> std::io::Result<Server> {
-    let inspect_path = config.inspect_path_prefix.clone() + "/{payload:.*}";
+    let inspect_base_path = config.inspect_path_prefix.clone();
```
gligneul, line 21: I think that this path doesn't need to be configurable anymore. Let's remove this configuration and simplify the code.

marcelstanley: Not sure about that, @gligneul. I have talked to @endersonmaia and having a configurable inspect path should be useful for load balancing purposes or something like that..

endersonmaia: @marcelstanley I had a conversation with @gligneul , and checking the current AWS ALB configuration, I think we can live with a hard-coded `/inspect` path.

I shall test this at AWS and give some feedback here if I find some problems.

torives (dismissed):

endersonmaia (approved):
Running at AWS and it works as expected.

marcelstanley:
> Running at AWS and it works as expected.

@endersonmaia, thanks!! But, just to clarify, are you referring to the POST implementation as a whole or to the [fixed `inspect` PATH](https://github.com/cartesi/rollups-node/pull/52#discussion_r1298279701)?

marcelstanley:
> Fixed `/inspect` path wasn't a problem, since it's now hardcoded into my deployments configuration. We should only take care of making it clear to document this somewhere (I know it's already on the OpenAPI Spec).


Thanks for clarifying, @endersonmaia!! I'll go ahead and remove the configurable inspect path from the code in another ticket.

@gligneul FYI

torives (approved):

PR #53, opened 2023-08-08
Remove claims from dispatcher

### LINKED ISSUE #40: Remove claims from the dispatcher
## 📚 Context

With the `authority-claimer` responsible for the claims, this section of the `dispatcher` code will be deprecated. The service that remains will need only to read the state and keep the machine updated.

## ✔️ Solution

Remove the redundant code from the `dispatcher`.

## 📈 Subtasks

- [x] Remove main loop calls
- [x] Remove blockchain driver
- [x] Remove obsolete auxiliary methods and tests
- [x] Rename `dispatcher` to a more appropriate name

GMKrieger:
#40 

Removed claims logic and auxiliary functions and utilities.

Renamed `dispatcher` to `eth-input-reader`.


renan061 (changes requested):
Minor comments!

The CI is broken, probably because of some name that still refers to "dispatcher".

GMKrieger:
> Minor comments!
> 
> The CI is broken, probably because of some name that still refers to "dispatcher".

I pushed the fix right before you commented, let's see if it works

GMKrieger:
Ok, so the check scopes is failing because the command we use to check is this

`git log --pretty=format:"%s" $(git merge-base HEAD origin/main)..HEAD`

Since this command compares to the main and we're currently branching on next, it's getting previous commits that are not from this branch that includes one with dispatcher as a scope, therefore causing this error.

There's no fix to be done as of right now, just putting this here so anyone that reviews doesn't get confused.

fmoura (dismissed):

renan061 (dismissed):

gligneul (changes requested):
Please don't merge this PR to `next/main` until the claimer is ready. Otherwise we won't be able to release a rollups version.

offchain/eth-input-reader/src/config.rs
```diff
@@ -19,12 +19,10 @@ use types::deployment_files::{
...
 };
 
-use crate::auth::{AuthConfig, AuthEnvCLIConfig, AuthError};
-
 #[derive(Clone, Parser)]
 #[command(name = "rd_config")]
-#[command(about = "Configuration for rollups dispatcher")]
-pub struct DispatcherEnvCLIConfig {
+#[command(about = "Configuration for rollups eth-input-reader")]
+pub struct EthInputReaderEnvCLIConfig {
```
gligneul, line 25: We still have the `TxEnvCLIConfig`  but the eth-input-reader doesn't send transactions. This seems wrong.

gligneul (changes requested):

offchain/eth-input-reader/src/config.rs
marcelstanley: @GMKrieger FYI

gligneul:
https://github.com/cartesi/rollups-node/issues/40#issuecomment-1815299765

PR #59, opened 2023-08-23
Update state-fold use on eth-input-reader

### LINKED ISSUE #41: Change how the state-fold is used in the dispatcher
## 📚 Context

The `dispatcher` currently uses the state-fold functionality through a grpc interface. With the state-fold now being a crate inside the workspace, it's better to change the implementation to use it directly.

## ✔️ Solution

Refactor the `dispatcher` code to use the state-fold directly, instead of using the grpc interface.

## 📈 Subtasks

- [x] Update state-fold calls
- [x] Update tests

GMKrieger:
#41


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Renamed `dispatcher` service to `eth-input-reader`
+- Refactored how `state-fold` is used inside `eth-input-reader`
```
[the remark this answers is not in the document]
gligneul: The main goal of the changelog is for the users to understand the changes so they can adapt their code. In our case, the main user is sunodo.

Here are some useful entries to the changelog:

1. Removed state-fold service
2. Renamed dispatcher service to eth-input-reader
3. Remove config variables X, Y, Z
4. Changed the name of config variables from A to B

We shouldn't mention internal changes that do not affect the user. E.g., refactors, internal modifications, etc.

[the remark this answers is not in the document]
torives: > The main goal of the changelog is for the users to understand the changes so they can adapt their code

I agree that's the main goal, but not the only one. As a contributor/developer, I find some refactors and internal modifications relevant enough to also be in the changelog.

I don't trust my memory, so if later on we find a bug in `eth-input-reader`, it would be helpful to look at the changelog and see that we ditched the `state-server` in this version and that might be a good starting point for troubleshooting.

I guess the real discussion is what is a "[notable change](https://keepachangelog.com/en/1.0.0/#what)". IMO this is notable enough.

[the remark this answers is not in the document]
gligneul: Yes, I agree, but it should be more descriptive, like you said. We can add multiple entries in the changelog as well.

torives (approved):

CHANGELOG.md
```diff
@@ -10,13 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added authority claimer service to support reader mode
+- Added `SFConfig`, `BHConfig` and `subscription_depth` configuration to `eth-input-reader`
```
gligneul, line 13: The internal names are not helpful for those who use the nome. We should add the full env var names here.

gligneul (changes requested):
I will do a full review tomorrow.

CHANGELOG.md
gligneul: Here is my proposal to the changelog:
```
### Changed
- Moved env vars A, B, C from state-fold config to eth-input-reader-config

### Removed
- Removed env vars A, B, C from eth-input-reader config
```

CHANGELOG.md
```diff
@@ -10,13 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Changed
 
 - Renamed `dispatcher` service to `eth-input-reader`
 
 ### Removed
```
gligneul, line 19: Let's remove the state-server from the repository in this PR because it won't be used anymore.

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,88 +1,108 @@
...
+    EthInputReaderError,
+> {
+    let env = StateFoldEnvironment::new(
+        provider,
+        Some(block_archive),
+        config.sf_config.safety_margin,
+        config.sf_config.genesis_block,
+        config.sf_config.query_limit_error_codes.clone(),
+        config.sf_config.concurrent_events_fetch,
+        10000,
```
gligneul, line 87: What is this magic number? We should at least create a local `const magic_number = 10000`.

gligneul (changes requested):

gligneul:
@GCdePaula, if you have some spare time, could you look at this PR?

offchain/eth-input-reader/src/config.rs
```diff
@@ -114,19 +122,20 @@ impl Config {
         let broker_config =
             BrokerConfig::from(eth_input_reader_config.broker_config);
 
-        assert!(
```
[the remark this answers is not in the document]
GMKrieger: Yes, this asserts compares the state client depth and transaction manager depth. The transaction manager has been removed from the project, so this assert is no longer necessary.

offchain/eth-input-reader/src/config.rs
```diff
@@ -43,25 +45,28 @@ pub struct EthInputReaderEnvCLIConfig {
...
     pub tx_config: TxManagerConfig,
     pub broker_config: BrokerConfig,
+    pub sf_config: SFConfig,
+    pub bh_config: BHConfig,
 
     pub dapp_deployment: DappDeployment,
     pub rollups_deployment: RollupsDeployment,
     pub epoch_duration: u64,
     pub priority: Priority,
+    pub subscription_depth: usize,
```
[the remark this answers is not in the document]
gligneul: Ok, it makes sense. Could you add this change to the backlog?

offchain/eth-input-reader/src/machine/driver.rs
```diff
@@ -5,7 +5,8 @@ use super::Context;
 
 use super::{rollups_broker::BrokerFacadeError, BrokerSend};
 
-use eth_state_fold_types::{ethereum_types::Address, Block};
+use eth_state_fold_types::Block;
+use ethers::types::Address;
```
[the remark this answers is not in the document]
GMKrieger: The type isn't changed, but before the state-fold-types re-exported the ethers crate, and we called the ethers types from inside state-fold-types. With the refactoring, we use state-fold-types only for the types defined in it, and import the ethers types directly from itself.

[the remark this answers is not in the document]
gligneul: The state-fold is an abstraction on top of ethers, so it makes sense to use the re-exported type instead of importing it from ethers directly.

offchain/eth-input-reader/src/setup.rs
GMKrieger: I got this magic number directly from state-server, where is just as magical, you can check on the state-server/src/lib.rs at line 92. I checked the StateFoldEnvironment and copied the comment from there above the const inside the setup.

offchain/eth-input-reader/src/eth_input_reader.rs
```diff
@@ -114,24 +136,21 @@ pub async fn start(
...
-
     broker: &(impl BrokerSend + BrokerReceive),
 ) -> Result<(), EthInputReaderError> {
     trace!("Querying rollup state");
-    let state = state_server
-        .query_state(initial_state, block.hash)
-        .await
-        .context(StateServerSnafu)?;
+    let state = RollupsState::get_state_for_block(
+        &Arc::new(initial_state.to_owned()),
```
GCdePaula, line 148: This line looks very weird to me. Can you verify?

offchain/eth-input-reader/src/eth_input_reader.rs
```diff
@@ -114,24 +136,21 @@ pub async fn start(
...
+        block,
+        env,
+    )
+    .await
+    .expect("should get state");
```
GCdePaula, line 153: At three arguments to just call this function (`initial_state`, `env`, `block`), maybe you should get the state before calling this function, and just pass the state.

I'd even rethink the role of this whole function (`process_block`). The main loop used to be on a different function too, separating it from setup. Now that they're joined, maybe there's a better way to do things.

I personally like smaller functions which do one think for readability and testability. YMMV. Whatever you prefer is fine, just evaluate if this balance is to your liking.

offchain/eth-input-reader/src/eth_input_reader.rs
```diff
@@ -49,10 +58,23 @@ pub async fn start(
...
+        Arc::clone(&provider),
+        Arc::clone(&block_subscriber.block_archive),
+    )
+    .await?;
+
     trace!("Creating context");
-    let mut context =
-        create_context(&config, &state_server, &broker, dapp_metadata, metrics)
-            .await?;
+    let mut context = create_context(
```
GCdePaula, line 70: This made sense when this service was doing two things (sending tx and reading state). I think the role of `context` may not make sense anymore.

Could you investigate whether it still makes sense? Maybe it should be part of the `machine_driver`.

offchain/eth-input-reader/src/eth_input_reader.rs
```diff
@@ -49,10 +58,23 @@ pub async fn start(
             .await
             .context(BrokerSnafu)?;
 
+    trace!("Creating env");
+    let env = create_env(
```
GCdePaula, line 62: I think the granularity of the setup function should be larger.

I'd create a function that returns both the `env` and the `block_subscriber`. The reason is because you don't need the `provider` nor the `subscriber`. Or at least do something like:

```
let (subscription, env) = {
   let provider = ...;
   let subscriber = ...;
};
```

offchain/eth-input-reader/src/eth_input_reader.rs
```diff
@@ -75,8 +97,8 @@ pub async fn start(
                     b.parent_hash
                 );
                 process_block(
+                    &Arc::clone(&env),
```
GCdePaula, line 100: This line looks weird to me. Questions:

Why is `env` wrapped in `Arc`? Why cloning it every loop iteration? If it's a reference, why clone before taking the reference?

offchain/eth-input-reader/src/eth_input_reader.rs
```diff
@@ -114,24 +136,21 @@ pub async fn start(
 #[instrument(level = "trace", skip_all)]
 #[allow(clippy::too_many_arguments)]
 async fn process_block(
+    env: &Arc<StateFoldEnvironment<InputProvider, Mutex<UserData>>>,
```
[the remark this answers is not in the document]
GCdePaula: Possibly just a reference instead of `Arc` _and_ reference?

offchain/eth-input-reader/src/machine/driver.rs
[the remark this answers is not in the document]
GCdePaula: Yeah, I'd be really cautious about not using reexported `ethers-rs`.

offchain/eth-input-reader/src/machine/driver.rs
```diff
@@ -5,7 +5,8 @@ use super::Context;
```
GCdePaula, line 5: I'd reconsider the name of `MachineDriver`. In retrospect, it's not amazing. And since there's no longer it's dual (`BlockchainDriver`), I'd try to think about something better.

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,88 +1,116 @@
...
 use snafu::ResultExt;
-use tokio_stream::{Stream, StreamExt};
-use tonic::transport::Channel;
-use types::foldables::authority::{RollupsInitialState, RollupsState};
+use std::sync::{Arc, Mutex};
+use tokio_stream::Stream;
+use types::UserData;
+use url::Url;
 
+use crate::error::{BlockArchiveSnafu, ParseSnafu};
```
GCdePaula, line 17: Put this inside the `crate::error` below.

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,88 +1,116 @@
...
-        .connect()
-        .await
-        .context(ConnectSnafu)?;
+pub type InputProvider = Provider<RetryClient<Http>>;
+
+pub async fn create_provider(
+    config: &EthInputReaderConfig,
+) -> Result<Arc<InputProvider>, EthInputReaderError> {
+    let http = Http::new(
+        Url::parse(&config.bh_config.http_endpoint).context(ParseSnafu)?,
```
GCdePaula, line 39: Ideally, this parsing should have been inside the `state-fold` crate. Anyway.

I think this is a configuration error (the user provided a bad url). I'd try to "lift" this check to the configuration crate. Also, if you do this, remove the `ParseError` variant from `crate::error::Error`.

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,88 +1,116 @@
...
+        config.sf_config.safety_margin,
+        config.sf_config.genesis_block,
+        config.sf_config.query_limit_error_codes.clone(),
+        config.sf_config.concurrent_events_fetch,
+        MAXIMUM_EVENTS_PER_RESPONSE,
+        Mutex::new(UserData::default()),
     );
 
-    Ok(s)
+    Ok(Arc::new(env))
```
GCdePaula, line 99: I'd reconsider the use of `Arc` here.

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,88 +1,116 @@
...
-pub async fn create_block_subscription(
-    client: &impl BlockServer,
-    confirmations: usize,
+pub async fn create_block_subscriber(
+    config: &EthInputReaderConfig,
+    provider: Arc<InputProvider>,
+) -> Result<Arc<BlockSubscriber<InputProvider>>, EthInputReaderError> {
+    Ok(Arc::new(
+        BlockSubscriber::start(
+            Arc::clone(&provider),
```
GCdePaula, line 56: Why clone?

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,88 +1,116 @@
...
 }
 
-pub async fn create_block_subscription(
-    client: &impl BlockServer,
-    confirmations: usize,
+pub async fn create_block_subscriber(
+    config: &EthInputReaderConfig,
+    provider: Arc<InputProvider>,
+) -> Result<Arc<BlockSubscriber<InputProvider>>, EthInputReaderError> {
+    Ok(Arc::new(
```
GCdePaula, line 54: I'd reconsider the use of `Arc` here.

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,88 +1,116 @@
...
     );
 
-    Ok(s)
+    Ok(Arc::new(env))
 }
 
 pub async fn create_context(
     config: &EthInputReaderConfig,
-    block_server: &impl BlockServer,
+    block_subscriber: Arc<BlockSubscriber<InputProvider>>,
```
GCdePaula, line 104: Possible just a reference: `&BlockSubscriber<InputProvider>`

offchain/eth-input-reader/src/eth_input_reader.rs
```diff
@@ -29,17 +34,21 @@ pub async fn start(
...
+    trace!("Creating block-subscriber");
+    let block_subscriber =
+        create_block_subscriber(&config, Arc::clone(&provider)).await?;
 
     trace!("Starting block subscription with confirmations");
-    let mut block_subscription = create_block_subscription(
-        &state_server,
-        config.sc_config.default_confirmations,
+    let mut block_subscription = create_subscription(
+        Arc::clone(&block_subscriber),
```
GCdePaula, line 50: Reconsider the need of this `clone`.

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,88 +1,116 @@
 // (c) Cartesi and individual authors (see AUTHORS)
 // SPDX-License-Identifier: Apache-2.0 (see LICENSE)
```
GCdePaula, line 3: I think many functions in this file could be joined into a single one, returning just what the `start` function needs.

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,88 +1,116 @@
...
     config: &EthInputReaderConfig,
-    block_server: &impl BlockServer,
+    block_subscriber: Arc<BlockSubscriber<InputProvider>>,
     broker: &impl BrokerStatus,
     dapp_metadata: DAppMetadata,
     metrics: EthInputReaderMetrics,
 ) -> Result<Context, EthInputReaderError> {
-    let genesis_timestamp: u64 = block_server
-        .query_block(config.dapp_deployment.deploy_block_hash)
+    let genesis_timestamp: u64 = block_subscriber
```
GCdePaula, line 109: I don't think this makes sense. I think you should use the provider instead.

offchain/log/src/lib.rs
```diff
@@ -10,7 +10,7 @@ pub mod built_info {
     include!(concat!(env!("OUT_DIR"), "/built.rs"));
 }
 
-#[derive(Debug, Parser)]
+#[derive(Debug, Parser, Clone)]
```
GCdePaula, line 13: I'd consider whether the cases you're cloning the config could instead just borrow.

offchain/rollups-events/src/broker/mod.rs
```diff
@@ -346,7 +346,7 @@ pub enum BrokerError {
     InvalidPayload { source: serde_json::Error },
 }
 
-#[derive(Debug, Parser)]
+#[derive(Debug, Parser, Clone)]
```
GCdePaula, line 349: I'd consider whether the cases you're cloning the config could instead just borrow.

GCdePaula (changes requested):
Nice!

I left a few comments. Also, I'm not sure what's the timeline here, and there's the Go node stuff. So maybe some of the comments are not really relevant. Feel free to ignore the minor and more abstract comments.

offchain/eth-input-reader/src/error.rs
```diff
@@ -29,8 +30,14 @@ pub enum EthInputReaderError {
...
 
-    #[snafu(display("state server error"))]
-    StateServerError { source: StateServerError },
+    #[snafu(display("parser error"))]
+    ParseError { source: ParseError },
+
+    #[snafu(display("parser error"))]
+    BlockArchiveError {
+        source:
+            eth_block_history::BlockArchiveError<Provider<RetryClient<Http>>>,
```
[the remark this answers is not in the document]
GMKrieger: I've trying doing it this way, but it conflicts with having `whatever` as part of the errors. If I remove the `whatever`, it works just fine. I tried looking at the documentation for a solution or reason, but couldn't find it.

offchain/eth-input-reader/src/error.rs
```diff
@@ -2,10 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0 (see LICENSE)
 
 use axum::http::uri::InvalidUri;
-use eth_state_client_lib::error::StateServerError;
+use ethers::providers::{Http, Provider, RetryClient};
```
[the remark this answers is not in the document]
GMKrieger: No good reason I can explain well. Renan was pretty adamant about not doing it this way, and at the time Rust would still be the main project going forward. I can change it back, no problem.

offchain/rollups-events/src/broker/mod.rs
GMKrieger: The borrow requires a lifetime specifier, the flatten command also complains, and the other configs from block-history and state-fold are cloned, so I just kept the pattern.

offchain/eth-input-reader/src/eth_input_reader.rs
GMKrieger: I investigated, and I think you're right. However, the change would be more trouble than it's worth, considering this is the last update to release the reader mode and will be deprecated eventually.

offchain/eth-input-reader/src/setup.rs
GMKrieger: How do I use the provider to get the genesis timestamp of a block?

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,90 +1,98 @@
...
-    impl Stream<Item = Result<BlockStreamItem, StateServerError>>
-        + Send
-        + std::marker::Unpin,
+    (
+        impl Stream<
+                Item = Result<
+                    BlockStreamItem,
+                    SubscriptionError<InputProvider>,
+                >,
+            > + std::marker::Unpin,
```
GCdePaula, line 46: This could be a type alias.

offchain/eth-input-reader/src/setup.rs
```diff
@@ -1,90 +1,98 @@
...
+        StateFoldEnvironment<InputProvider, Mutex<UserData>>,
```
GCdePaula, line 47: Maybe this too?

GCdePaula (approved):
LGTM. Just some minor comments on style.

Issue #7, opened 2023-04-20
Standardize node config


gligneul:
## 📚 Context

Each service has its own way of getting config values.

## ✔️ Solution

Change the config to be standardized across all services.

## 📈 Subtasks

- [ ] Migrate from struct-opt to clap. Clap merged the struct-opt features and made it deprecated.
- [ ] Improve the documentation and default values of the config variables. In particular, we should improve the documentation of variables with abbreviations (e.g. HC, RD, SC, and TX).
- [ ] Move all config to a single file in the root of each crate. Some crates have their configs hidden in the middle of the code.
- [ ] Standardize the error handling in the config. I'm in favor of using expect instead of errors because config errors are not recoverable anyway.

gligneul:
Replaced by https://github.com/cartesi/rollups-node/issues/88

PR #61, opened 2023-08-23
feat(advance-runner)!: validate snapshot hash

### LINKED ISSUE #36: Node should check the machine hash from blockchain

torives:
The snapshot's template hash gets compared with its counterpart in the blockchain, causing the advance-runner to stop if it does not match.

BREAKING CHANGE: adds the PROVIDER_HTTP_ENDPOINT parameter, which is required to validate snapshots. Validation is enabled by default, but can be disabled by setting SNAPSHOT_VALIDATION_ENABLED to false

Closes #36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added authority claimer service to support reader mode
 - Added support to `POST` *inspect state* requests
+- Added snapshot validation. The node will now check whether the snapshot's template hash matches the one stored in the blockchain
 
 ### Changed
 
--- a/offchain/advance-runner/Cargo.toml
+++ b/offchain/advance-runner/Cargo.toml
@@ -9,6 +9,7 @@ name = "cartesi-rollups-advance-runner"
 path = "src/main.rs"
 
 [dependencies]
+contracts = { path = "../contracts" }
 grpc-interfaces = { path = "../grpc-interfaces" }
 http-health-check = { path = "../http-health-check" }
 log = { path = "../log" }
@@ -17,13 +18,15 @@ rollups-events = { path = "../rollups-events" }
 async-trait.workspace = true
 backoff = { workspace = true, features = ["tokio"] }
 clap = { workspace = true, features = ["derive", "env"] }
+ethers.workspace = true
 hex.workspace = true
 sha3 = { workspace = true, features = ["std"] }
 snafu.workspace = true
 tokio = { workspace = true, features = ["macros", "time", "rt-multi-thread"] }
 tonic.workspace = true
-tracing.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
+tracing.workspace = true
+url.workspace = true
 uuid = { workspace = true, features = ["v4"] }
 
 [dev-dependencies]
@@ -32,5 +35,5 @@ test-fixtures = { path = "../test-fixtures" }
 env_logger.workspace = true
 rand.workspace = true
 tempfile.workspace = true
-testcontainers.workspace = true
 test-log = { workspace = true, features = ["trace"] }
+testcontainers.workspace = true
--- a/offchain/advance-runner/src/dapp_contract.rs
+++ b/offchain/advance-runner/src/dapp_contract.rs
@@ -1,0 +1,44 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+use contracts::cartesi_dapp::CartesiDApp;
+use ethers::{
+    prelude::ContractError,
+    providers::{Http, HttpRateLimitRetryPolicy, Provider, RetryClient},
+};
+use rollups_events::{Address, Hash};
+use snafu::{ResultExt, Snafu};
+use std::sync::Arc;
+use url::Url;
+
+const MAX_RETRIES: u32 = 10;
+const INITIAL_BACKOFF: u64 = 1000;
+
+#[derive(Debug, Snafu)]
+#[snafu(display("failed to obtain hash from dapp contract"))]
+pub struct DappContractError {
+    source: ContractError<Provider<RetryClient<Http>>>,
+}
+
+pub async fn get_template_hash(
+    dapp_address: &Address,
+    provider_http_endpoint: Url,
+) -> Result<Hash, DappContractError> {
+    let provider = Provider::new(RetryClient::new(
+        Http::new(provider_http_endpoint),
+        Box::new(HttpRateLimitRetryPolicy),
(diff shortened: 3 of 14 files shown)


endersonmaia:
Since you're already adding a BREAKING CHANGE with `PROVIDER_HTTP_ENDPOINT` parameter, I think it's a great opportunity to standardize the PROVIDER parameter across services.

**dispatcher**

    TX_PROVIDER_HTTP_ENDPOINT => PROVIDER_HTTP_ENDPOINT

**state-server**

    BH_HTTP_ENDPOINT => PROVIDER_HTTP_ENDPOINT
    BH_WS_ENDPOINT => PROVIDER_WS_ENDPOINT

gligneul:
@endersonmaia, these changes are unrelated to this PR. Also, they are in another service and involve changing external dependencies (the state fold and the transaction manager). So we should not do this now.

endersonmaia:
Should I open an issue for that?

gligneul:
Here is the related issue: https://github.com/cartesi/rollups-node/issues/7

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Added
 
 - Added authority claimer service to support reader mode
+- Added snapshot validation. The node will now check if the snapshot's template hash matches the one stored in the blockchain
```
gligneul, line 13:
```suggestion
- Added snapshot validation. The node will now check whether the snapshot's template hash matches the one stored in the blockchain
```

marcelstanley (commented):
LGTM

marcelstanley (dismissed):

gligneul (changes requested):
Will review

offchain/advance-runner/Cargo.toml
```diff
@@ -16,13 +16,16 @@ rollups-events = { path = "../rollups-events" }
 async-trait.workspace = true
 backoff = { workspace = true, features = ["tokio"] }
 clap = { workspace = true, features = ["derive", "env"] }
+contracts = { path = "../contracts" }
```
gligneul, line 19: This should be with the other local dependencies (grpc-interfaces, http-healthcheck, rollups-events).

offchain/advance-runner/src/dapp_contract.rs
```diff
@@ -0,0 +1,46 @@
...
+use rollups_events::{Address, Hash};
+use snafu::{ResultExt, Snafu};
+use std::sync::Arc;
+use url::Url;
+
+const MAX_RETRIES: u32 = 10;
+const INITIAL_BACKOFF: u64 = 1000;
+
+#[derive(Debug, Snafu)]
+pub enum DappContractError {
```
gligneul, line 18: Since this has only one options it should be a struct.

offchain/advance-runner/src/snapshot/fs_manager.rs
```diff
@@ -213,26 +221,26 @@ impl SnapshotManager for FSSnapshotManager {
...
-            .read(&mut buffer)
-            .context(ReadHashSnafu { path: path.clone() })?;
-        tracing::trace!("read {bytes} bytes from file");
-        Ok(Hash::new(buffer))
+            match offchain_hash == onchain_hash {
+                true => Ok(()),
+                false => Err(FSSnapshotError::HashMismatchError),
+            }
+        } else {
+            tracing::info!("snapshot validation disabled, accepting snapshot");
```
gligneul, line 241:
```suggestion
            tracing::trace!("snapshot validation disabled, accepting snapshot");
```

offchain/advance-runner/src/snapshot/fs_manager.rs
```diff
@@ -213,26 +221,26 @@ impl SnapshotManager for FSSnapshotManager {
...
+            .context(OnchainSnafu)?;
 
-        let mut buffer = [0_u8; HASH_SIZE];
-        let bytes = file
-            .take(HASH_SIZE as u64)
-            .read(&mut buffer)
-            .context(ReadHashSnafu { path: path.clone() })?;
-        tracing::trace!("read {bytes} bytes from file");
-        Ok(Hash::new(buffer))
+            match offchain_hash == onchain_hash {
```
gligneul, line 236: Match true-false seems very strange. I'd instead use if-else.

offchain/advance-runner/src/dapp_contract.rs
```diff
@@ -0,0 +1,46 @@
...
+    #[snafu(display("failed to obtain hash from dapp contract"))]
+    ContractError {
+        source: ContractError<Provider<RetryClient<Http>>>,
+    },
+}
+
+pub async fn get_template_hash(
+    dapp_address: Address,
+    provider_http_endpoint: Url,
```
gligneul, line 27: Couldn't these be references?

offchain/advance-runner/src/snapshot/fs_manager.rs
gligneul, line 236: This will only succeed if the advance runner loads a snapshot that is the first one.

gligneul (changes requested):

offchain/advance-runner/src/dapp_contract.rs
torives: `dapp_address ` can be a reference no problem. But `Url` will be consumed in the function. Is it better/more idiomatic to clone it before passing it as a parameter or cloning the reference inside?

offchain/advance-runner/src/snapshot/fs_manager.rs
torives: Completely overlooked this "detail". Thanks for catching it! 😉

offchain/advance-runner/src/dapp_contract.rs
gligneul: If it is consumed, it is better to receive a value.

gligneul (dismissed):

gligneul (approved):

PR #63, opened 2023-08-25
ci: check if PR updated changelog

torives:
Add a new CI workflow that checks whether a PR targeting the main branch modified the CHANGELOG.md file. This can be disabled by using the "no changelog" tag when opening the PR.

To test this you can remove the tag from this PR. The workflow should run again and fail.

### DIFF
--- a/.github/workflows/changelog-check.yml
+++ b/.github/workflows/changelog-check.yml
@@ -1,0 +1,18 @@
+name: Check Changelog
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches:
+      - main
+
+jobs:
+  check-changelog:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: tarides/changelog-check-action@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog-check.yml
```diff
@@ -0,0 +1,18 @@
+name: Check Changelog
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
gligneul, line 5: Do we need this?

torives: The labeled/unlabeled is required for the opt-out feature. I'm checking the docs to see if the other 3 are necessary.

torives: "opened", "synchronize" and "reopened" are the [default](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows) for a "pull_request". But since we need to specify "labeled" and "unlabeled", we need to add them too otherwise the workflow won't trigger in those situations.

gligneul (approved):
Nice

PR #87, opened 2023-09-14
fix: Fix help generation

fmoura:


offchain/advance-runner/src/config.rs
```diff
@@ -62,7 +62,9 @@ pub enum ConfigError {
     SnapshotConfigError { source: SnapshotConfigError },
 }
 
-#[derive(Parser, Debug)]
+#[derive(Clone, Parser)]
+#[command(name = "advance_runner_config")]
+#[command(about = "Configuration for rollups advance-runner")]
```
gligneul, line 67: I prefer that we remove the about section. Most CLI starts with the usage.

offchain/advance-runner/src/config.rs
```diff
@@ -62,7 +62,9 @@ pub enum ConfigError {
     SnapshotConfigError { source: SnapshotConfigError },
 }
 
-#[derive(Parser, Debug)]
+#[derive(Clone, Parser)]
```
gligneul, line 65: Do we need the clone?

gligneul (changes requested):

offchain/advance-runner/src/config.rs
fmoura: Some use case needed, not sure why. I was just making it standard. Will remove and re-check who really needs

gligneul (changes requested):
The help message still needs to be clarified.
```
➜  offchain git:(683378c8) ./target/debug/cartesi-rollups-advance-runner --help | head
CLI configuration used to generate the DApp metadata

Usage: cartesi-rollups-advance-runner [OPTIONS] --snapshot-dir <SNAPSHOT_DIR> --snapshot-latest <SNAPSHOT_LATEST>
```

If it is too hard to remove the description, we could just leave something like "Configuration for advance-runner".
I didn't test the other services.

fmoura:
> CLI configuration used to generate the DApp metadata

This one comes from a Comment and not an annotation. Maybe it is best to add back again the Description for every service, so we do not fallback on presenting some unwanted description that comes from a "inner" configuration that could come from an external crate

fmoura:
Correctly tested every service and they are consistent now, looking something like this
```
docker run -it cartesi/rollups-node:devel cartesi-rollups-state-server --help | head
Configuration for state-server

Usage: cartesi-rollups-state-server [OPTIONS]
```

gligneul (approved):

PR #89, opened 2023-09-15
feat: add node simplified config

### LINKED ISSUE #88: Simplify Node config
## 📚 Context

The node configuration is confusing because it is spread across multiple services.
It doesn't follow a single naming convention, and we have different variables with the same meaning.

## ✔️ Solution

We want to provide a set of environment variables in the new rollups-node monolith to help configure it. We will concentrate the documentation for these variables in a single place so it is easy to reason. We will also merge variables with the same meaning but different names. For instance, BH_HTTP_ENDPOINT and TX_PROVIDER_HTTP_ENDPOINT will become CARTESI_PROVIDER_HTTP_ENDPOINT.

We plan to add the CARTESI prefix to all variables to avoid clashing with previous names. If an internal variable is already set, the monolith will not overwrite it. For example, if you set BH_HTTP_ENDPOINT, the monolith won't overwrite it with CARTESI_PROVIDER_HTTP_ENDPOINT. Notice, however, that the monolith will set the TX_PROVIDER_HTTP_ENDPOINT to CARTESI_PROVIDER_HTTP_ENDPOINT if it isn't set.

We should define safe default values as much as possible to start the node defining with minimal configuration. Variables that can be set directly from the monolith won't be exposed. For instance, the address of the state server (SC_GRPC_ENDPOINT) will be hardcoded by the monolith to a local host address. If SC_GRPC_ENDPOINT is already set, the monolith won't overwrite, just like the case above. However, this is not recommended.

Here is the list of variables we will provide and where they are used. (In the final version, we won't document where they are used; this is just for development tracking.)

### CARTESI_LOG_LEVEL

Sets RUST_LOG in all node services.
We have to think about the server manager log; currently, it is defined by:
- SERVER_MANAGER_LOG_LEVEL=warning
- REMOTE_CARTESI_MACHINE_LOG_LEVEL=info

### CARTESI_GRAPHQL_PORT

- graphql-server: GRAPHQL_PORT

### CARTESI_INSPECT_PORT

- inspect-server: INSPECT_SERVER_ADDRESS

In this case, INSPECT_SERVER_ADDRESS=0.0.0.0:$CARTESI_INSPECT_PORT

### CARTESI_POSTGRES_*

- indexer and graphql-server: POSTGRES_*

### CARTESI_REDIS_ENDPOINT

- dispatcher, advance-runner, indexer: REDIS_ENDPOINT

### CARTESI_PROVIDER_HTTP_ENDPOINT

- dispatcher: TX_PROVIDER_HTTP_ENDPOINT
- state_server: BH_HTTP_ENDPOINT
- advance-runner: PROVIDER_HTTP_ENDPOINT

### CARTESI_PROVIDER_WS_ENDPOINT

- state-server: BH_WS_ENDPOINT

### CARTESI_DAPP_DEPLOYMENT_FILE

- dispatcher: RD_DAPP_DEPLOYMENT_FILE
- advance-runner and indexer: DAPP_CONTRACT_ADDRESS_FILE

### CARTESI_ROLLUPS_DEPLOYMENT_FILE

- dispatcher: RD_ROLLUPS_DEPLOYMENT_FILE

### CARTESI_CHAIN_ID

- dispatcher: TX_CHAIN_ID
- advance-runner and indexer: CHAIN_ID

### CARTESI_CHAIN_IS_LEGACY

- dispatcher: TX_CHAIN_IS_LEGACY

### CARTESI_EPOCH_DURATION

- dispatcher: RD_EPOCH_DURATION

### CARTESI_STATE_FOLD_GENESIS_BLOCK

- state-server: SF_GENESIS_BLOCK

### CARTESI_STATE_FOLD_SAFETY_MARGIN

- state-server: SF_SAFETY_MARGIN

### CARTESI_STATE_FOLD_BLOCK_TIMEOUT

- state-server: BH_BLOCK_TIMEOUT

### CARTESI_STATE_FOLD_DEFAULT_CONFIRMATIONS

- dispatcher: SC_DEFAULT_CONFIRMATIONS

### CARTESI_TRANSACTION_DEFAULT_CONFIRMATIONS

Can we merge this one with CARTESI_STATE_SERVER_DEFAULT_CONFIRMATIONS?

- dispatcher: TX_DEFAULT_CONFIRMATIONS

### CARTESI_AUTH_*

- dispatcher: AUTH_*

### CARTESI_SERVER_MANAGER_SESSION_ID

- advance_runner and inspect_server: SESSION_ID

### CARTESI_SNAPSHOT_DIR

- advance_runner: SNAPSHOT_DIR

We can set SNAPSHOT_LATEST in advance_runner to $SNAPSHOT_DIR/latest

### CARTESI_SNAPSHOT_ENABLED

- advance_runner: SNAPSHOT_ENABLED

gligneul:
Close #88

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/cartesi/rollups-node
+
+go 1.21.1
--- a/internal/cmd/print_config/main.go
+++ b/internal/cmd/print_config/main.go
@@ -1,0 +1,10 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package main
+
+import "github.com/cartesi/rollups-node/internal/pkg/config"
+
+func main() {
+	config.PrintDocumentation()
+}
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -1,0 +1,35 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+// This package loads the node config from enviroment variables
+package config
+
+import (
+	"fmt"
+	"os"
+)
+
+// Prefix to all enviroment variables
+const Prefix string = "CARTESI_"
+
+// Definition of config variables.
+// For each variable declared below, there should be a entry in meta.go.
+
+var PrintConfig bool
+var LogLevel string
+
+// Prints the documentation for each variable in Stdout
+func PrintDocumentation() {
+	genDocumentation(os.Stdout)
+}
+
+// Loads the config from the environment.
+// If PrintConfig is true, prints the config to Stdout.
+// This function calls os.Exit if there is an error.
+func Load() {
+	err := loadFromEnv(os.Stdout)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
--- a/internal/pkg/config/meta.go
+++ b/internal/pkg/config/meta.go
@@ -1,0 +1,78 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package config
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"slices"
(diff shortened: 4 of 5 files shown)


gligneul:
I won't follow this approach. I will create a new PR later.

PR #93, opened 2023-09-19
refactor: remove conventional commits scope check

### LINKED ISSUE #92: Remove conventional commits scope check
## 📚 Context

In the past, we decided to enforce a fixed set of scopes for conventional commits.
After experimenting with these for a release cycle, we decided that enforcing the scopes was not worth the extra hassle.

## ✔️ Solution

We should remove the scope check from the CI, the .scopes, and the .check_scopes.lua files, and update the READ-ME accordingly.

gligneul:
Close #92


gligneul:
I'm not adding this to the changelog on purpose because this is a small internal change.

GMKrieger (approved):

PR #94, opened 2023-09-19
feat: build single node binary

### LINKED ISSUE #91: Create the base Go project
## 📚 Context

This is the starting point for implementing #80, with the goal of exercising its main functions and APIs. The `graphql-server` was chosen as the first service for being simpler and therefore easier to test.

## 📈 Subtasks

- [x] Handle the execution mode CLI parameter
- [x] Forward the environment variables
- [x] Start `graphql-server`
- [x] Handle stdout and stderr

torives:
### Setup the database
The binary will start the `graphql-server` service and therefore needs a PostgreSQL instance to connect to. You should follow the instructions in the `data` crate [README](https://github.com/cartesi/rollups-node/blob/feature/single-binary/offchain/data/README.md) file to get a populated database you can interact with.

### Build `graphql-server`
The `cartesi-node` command assumes the executables (`graphql-server` in this case) are present in the PATH variable. So you'll first need to build it with cargo and then add it to your PATH. For example:

Inside `offchain/graphql-server`:
```sh
cargo build
```
and then:
```sh
PATH="{path to your workspace}/rollups-node/offchain/target/debug:${PATH}" 
```
### Run the `cartesi-node` command
From the project's root, just run:
```sh
POSTGRES_PASSWORD=pw go run ./cmd/cartesi-node full
```
The server should have started and to verify it you can interact with it via the GraphQL Playground at `http://localhost:4000/graphql`

Closes #91

### DIFF
--- a/cmd/cartesi-node/full.go
+++ b/cmd/cartesi-node/full.go
@@ -1,0 +1,52 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package main
+
+import (
+	"io"
+	"log"
+	"os"
+	"os/exec"
+
+	"github.com/spf13/cobra"
+)
+
+var full = &cobra.Command{
+	Use:   "full",
+	Short: "Starts the node in full mode with reader and validator capabilities",
+	Run:   runFullNode,
+}
+
+func runFullNode(cmd *cobra.Command, args []string) {
+	proc := exec.Command("cartesi-rollups-graphql-server")
+
+	stdoutPipe, err := proc.StdoutPipe()
+	if err != nil {
+		log.Fatal("Error creating stdout pipe:", err)
+	}
+
+	stderrPipe, err := proc.StderrPipe()
+	if err != nil {
+		log.Fatal("Error creating stderr pipe:", err)
+	}
+
+	if err := proc.Start(); err != nil {
+		log.Fatal("Error starting sub-process:", err)
+	}
+
+	// Create goroutines to display the sub-process's stdout and stderr.
+	go func() {
+		io.Copy(os.Stdout, stdoutPipe)
+	}()
+
+	go func() {
+		io.Copy(os.Stderr, stderrPipe)
+	}()
+
+	if err := proc.Wait(); err != nil {
+		log.Fatal("Error waiting for sub-process:", err)
+	}
+
+	log.Println("Sub-process finished")
+}
--- a/cmd/cartesi-node/main.go
+++ b/cmd/cartesi-node/main.go
@@ -1,0 +1,14 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package main
+
+import (
+	"os"
+)
+
+func main() {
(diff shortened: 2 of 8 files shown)


cmd/supervisor/supervisor.go
```diff
@@ -0,0 +1,16 @@
+// (c) Cartesi and individual authors (see AUTHORS)
```
gligneul, line 1: Let's rename this package. I don't want to give the wrong impression with the supervisor name. Probably, it is best to stick with the name of the final binary. Something like `rollups-node`, `cartesi-node`, or `cartesi-rollups-node`.

internal/pkg/supervisor/full.go
```diff
@@ -0,0 +1,52 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package supervisor
```
gligneul, line 4: This sub-package is redundant. The cmd package has nothing else, so we should move all these files to `cmd/<binary-name>/`. Let's start simple. When we do so, we can use internal symbols for everything (symbols that start with a lowercase).

cmd/supervisor/supervisor.go
```diff
@@ -0,0 +1,16 @@
...
+
+package main
+
+import (
+	"os"
+
+	cmd "github.com/cartesi/rollups-node/internal/pkg/supervisor"
+)
+
+func main() {
```
gligneul, line 12: It would be clearer if the file containing the main function is called `main.go`.

internal/pkg/supervisor/full.go
```diff
@@ -0,0 +1,52 @@
...
+		log.Fatal("Error starting sub-process:", err)
+	}
+
+	// Create goroutines to display the sub-process's stdout and stderr.
+	go func() {
+		io.Copy(os.Stdout, stdoutPipe)
+	}()
+
+	go func() {
+		io.Copy(os.Stdout, stderrPipe)
```
gligneul, line 44:
```suggestion
		io.Copy(os.Stderr, stderrPipe)
```

internal/pkg/supervisor/host.go
```diff
@@ -0,0 +1,14 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package supervisor
+
+import "github.com/spf13/cobra"
+
+var host = &cobra.Command{
```
gligneul, line 8: We should rename host mode to `no-backend` to keep in sync with sunodo.

internal/pkg/supervisor/root.go
```diff
@@ -0,0 +1,18 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package supervisor
+
+import "github.com/spf13/cobra"
+
+var Root = &cobra.Command{
+	Use: "cartesi-node-supervisor [reader|validator|full|host]",
```
gligneul, line 9: Rename `host` here as well.

gligneul (changes requested):

gligneul (approved):

PR #104, opened 2023-09-28
Use goroutines to start services

### LINKED ISSUE #96: Run services using goroutines
## 📚 Context
Adding new services to the `cartesi-node` binary requires calling blocking operations. We must avoid blocking the main thread while still managing the lifecycle and output of multiple sub-processes.

## ✔️ Solution
Run services inside goroutines and use channels to communicate with them. An implementation was suggested by @gligneul [here](https://github.com/cartesi/rollups-node/issues/80#issuecomment-1725764657)

torives:
Closes #96


cmd/cartesi-node/root.go
```diff
@@ -3,16 +3,21 @@
...
+
+	"github.com/spf13/cobra"
+)
 
 var rootCmd = &cobra.Command{
 	Use: "cartesi-node [reader|validator|full|no-backend]",
 	Run: func(cmd *cobra.Command, args []string) { cmd.Usage() },
 }
 
+const DefaultServiceTimeout = 1 * time.Minute
```
gligneul, line 17: This seems out of place.

internal/pkg/services/service.go
```diff
@@ -0,0 +1,16 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package services
```
gligneul, line 4: We should add some documentation to the package.

cmd/cartesi-node/root.go
torives: I wasn't really happy with it either. Accepting suggestions 😄

torives: I'll leave it in the file it is used, for now at least.

internal/pkg/services/graphql-service.go
```diff
@@ -0,0 +1,42 @@
...
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+const serviceName = "cartesi-rollups-graphql-server"
+
+type GraphQLService struct{}
+
+func (g *GraphQLService) Start(ctx context.Context) error {
```
gligneul, line 18:
```suggestion
func (g GraphQLService) Start(ctx context.Context) error {
```

This doesn't need to be a pointer.

cmd/cartesi-node/root.go
```diff
@@ -3,16 +3,17 @@
...
 package main
 
-import "github.com/spf13/cobra"
+import (
+	"github.com/spf13/cobra"
+)
 
 var rootCmd = &cobra.Command{
-	Use: "cartesi-node [reader|validator|full|no-backend]",
+	Use: "cartesi-node [reader|validator|no-backend]",
```
gligneul, line 11: Not related to this line. I think we can reword the commit message to:
```
feat: add graphql-server to the rollups-node binary
```

internal/pkg/services/graphql-service.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+const serviceName = "cartesi-rollups-graphql-server"
```
gligneul, line 14:
```suggestion
const (
    serviceName = "graphql-server"
    binaryName = "cartesi-rollups-graphql-server"
)
```

gligneul (changes requested):

cmd/cartesi-node/root.go
gligneul, line 11: The usage seems strange:
```
Usage:
  cartesi-node [reader|validator|no-backend] [flags]
  cartesi-node [command]
```

cmd/cartesi-node/validator.go
```diff
@@ -3,12 +3,21 @@
...
 	Short: "Starts the node in validator mode",
-	Run: func(cmd *cobra.Command, args []string) {
-		println("TODO")
-	},
+	Run:   runValidatorNode,
+}
+
+func runValidatorNode(cmd *cobra.Command, args []string) {
+	validatorServices := []services.Service{
+		&services.GraphQLService{},
```
gligneul, line 19:
```suggestion
		services.GraphQLService{},
```
We don't need to use the GraphQLService as a pointer.

internal/pkg/services/service.go
```diff
@@ -0,0 +1,70 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+// Contains the several internal services that comprises the cartesi-node and
+// the interfaces and functions used to interact with them
```
gligneul, line 5:
```suggestion
// Provide mechanisms to start multiple services in the background
```

internal/pkg/services/service.go
```diff
@@ -0,0 +1,70 @@
...
+		cancel()
+		for i := 0; i < len(services)-1; i++ {
+			<-exit
+		}
+		wait <- struct{}{}
+	}()
+
+	select {
+	case <-wait:
+		fmt.Println("main: all services exited")
```
gligneul, line 66: We should start to use log instead of fmt.Println

internal/pkg/services/service.go
```diff
@@ -0,0 +1,70 @@
...
+// Defines the interface of a service
+type Service interface {
+	fmt.Stringer
+
+	// Start a service that will run until completion or until the context is
+	// canceled
+	Start(ctx context.Context) error
+}
+
+const DefaultServiceTimeout = 1 * time.Minute
```
gligneul, line 23: I would start with 15 seconds.

internal/pkg/services/service.go
```diff
@@ -0,0 +1,70 @@
...
+		for i := 0; i < len(services)-1; i++ {
+			<-exit
+		}
+		wait <- struct{}{}
+	}()
+
+	select {
+	case <-wait:
+		fmt.Println("main: all services exited")
+	case <-time.After(DefaultServiceTimeout):
```
gligneul, line 67: nice

internal/pkg/services/service.go
torives: That's better 😄

internal/pkg/services/service.go
torives: I agree. I guess we should configure it in the same way we did in the Rust code. I'll update the PR as soon as I figure out how to do it.

torives: The `log` package from the standard library is quite simple, it doesn't support different levels or changing the date format. I'll create another issue for this.

cmd/cartesi-node/root.go
torives: I've tweaked it a little bit:

```
Usage:
  cartesi-node [command]

Available Commands:
  help        Help about any command
  no-backend  Starts the node in no-backend mode
  reader      Starts the node in reader mode
  validator   Starts the node in validator mode

Flags:
  -h, --help   help for cartesi-node

Use "cartesi-node [command] --help" for more information about a command.
```

gligneul (dismissed):
Looks good!

gligneul (approved):

internal/pkg/services/graphql-service.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	go func() {
+		<-ctx.Done()
+		fmt.Printf("%v: %v\n", g.String(), ctx.Err())
+		cmd.Process.Signal(syscall.SIGTERM)
+	}()
+
+	err := cmd.Wait()
+	if err != nil && cmd.ProcessState.ExitCode() != int(syscall.SIGTERM) {
```
renan061, line 37: Shouldn't this && be and assert? What if the ExitCode is not different but err is not nil?

renan061 (approved):

internal/pkg/services/graphql-service.go
gligneul: I don't like asserting here because this is an external process. I'd rather return the error and leave the decision to the callee (probably just log the error).

torives: There's the edge case where the process ends because of a `SIGTERM` that we didn't trigger. But considering that this is a temporary setup and will be thrown away after we fully rebuild the node in Go, it shouldn't be a significant concern.

Issue #130, opened 2023-11-06
Make dispatcher work with authority-claimer


gligneul:
## 📚 Context

For the 1.2 release, the `dispatcher` must stop doing the claims to allow the `authority-claimer` to be the sole responsible for the claims in the Node.

## ✔️ Solution

Remove the claims code from the `dispatcher`.

## 📈 Subtasks

- [ ] Remove deprecated configs
- [ ] Remove claims logic
- [ ] Remove obsolete auxiliary methods and tests

gligneul:
Closed by https://github.com/cartesi/rollups-node/pull/135

PR #135, opened 2023-11-06
Authority claimer with dispatcher for release 1.2

GMKrieger:
#130


CHANGELOG.md
```diff
@@ -11,11 +11,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `cartesi-rollups-node` Go binary as a single entrypoint to execute all Cartesi Node services
 - Added `authority-claimer` service
```
marcelstanley, line 13: Perhaps we should make it clear that the dispatcher no longer sends claims

marcelstanley (commented):
Looks good! Other than the comment below, I believe we could add a `!` to the commit message as well

offchain/authority-claimer/src/checker.rs
```diff
@@ -88,7 +88,7 @@ impl DuplicateChecker for DefaultDuplicateChecker {
         rollups_claim: &RollupsClaim,
     ) -> Result<bool, Self::Error> {
         let proof_context =
-            U256([0, 0, 0, rollups_claim.epoch_index]).encode().into();
```
gligneul, line 91: Is this change intended? If yes, please add it in another commit.

gligneul:
Also, let's test this on rollups-examples before merging. I want to avoid breaking the main branch by accident.

offchain/authority-claimer/src/checker.rs
GMKrieger: It's intentional. The way it was before, it was building the proof hex the other way around, e.g. 0x01000000 instead of 0x00000001. The claim contracts were not happy with this, as both the get couldn't find it and submit thought we skipped epochs.

marcelstanley (approved):

Issue #80, opened 2023-09-08
Create Go supervisor


torives:
## 📚 Context

Running a node right now requires the orchestration of several Docker containers. For testing, running locally, and/or deploying in a VPS a simpler option is desirable.

Original issue/discussion: #26

## ✔️ Solution

Create a single binary to execute all of our services. All service configuration should be done via environment variables, as the binary will have just a single parameter defining its execution mode: reader, validator, full (reader+validator), or host.

It will forward each service's stdout and stderr to its own stdout and stderr, similar to what is currently done when running the node using the `docker-compose.yml` file.

Although `server-manager` is not a service built in this repository, it will also be executed by the binary: it will assume it is installed in the system and available in the PATH.

It will also assume the Redis, PostgreSQL, and Hardhat services are already running and ready to connect by the time it starts.

## 📈 Subtasks

- [x] #65
- [x] #91
- [x] #96
- [x] #108
- [x] #116
- [x] #88
- [x] #125
- [x] #97
- [x] #139
- [x] #140
- [x] #180
- [x] #179
- [x] #177
- [x] #193
- [x] #221
- [x] #241
  - `/graphql`, `/inspect`, `/rollups` (host-mode), `/metrics` (dispatcher)
  - See https://github.com/elazarl/goproxy
  - We should expose only the proxy port once we have the proxy. Addresses other than the port should be 127.0.0.1.
  - The proxy address should be 127.0.0.1 by default, and we should have an env var to configure it. This way, the node doesn't bind to a public address, which would be an unsafe default.
- [x] #198
- [x] #222
  - See: https://github.com/gligneul/eggroll/blob/main/internal/dev/signal.go#L14 and https://github.com/gligneul/eggroll/blob/main/internal/dev/dev.go#L27
- [x] #200

gligneul:
We could rename `host` mode to `backendless` mode, following the idea of sunodo.

endersonmaia:
> Create a single binary to execute all of our services....

> It will forward each service's stdout and stderr to its own stdout and stderr ...

Are you planning to create another process/service supervisor? Why not reuse some existing software that already does this?

For sunodo/rollups-node, we're using [s6-overlay,](https://github.com/just-containers/s6-overlay), but there are a lot of other options if you think `s6` is not a fit.

The choice for `s6` was 'cause it's made for running inside containers and it's very light.

endersonmaia:
[...]

Issue #29, opened 2023-05-17
Add infra for manual tests


gligneul:
## 📚 Context

We use rollup examples for manual node running and testing.
Testing the node with infrastructure from a different repository significantly reduces team efficiency.

## ✔️ Solution

We should add infrastructure to run the node in this repository.
The main objective of this infrastructure is for the node unit to test and debug the node.
Having a generic solution to run any DApp in the node is not an objective.

## 📈 Subtasks

- [x] #173
- [x] #149
- [x] #187
- [x] #213
    - We should create this directory in the host system to facilitate running the node without docker.
- [x] #217
    - Anvil and postgres containers.
    - We should consider using testcontainers to manage the docker containers, so this code can be reused later when we implement automatic e2e tests.
- [x] Add command to run the node with docker
    - The node should connect to the dependencies from the previous command.
    - We need a local volume to map the snapshot dir.
- [x] #264
    - Include section on how to run the node with docker.
- [x] #232
    - Include section on how to run the node without docker.

gligneul:
Done

PR #138, opened 2023-11-07
feat: add files to support manual tests

### LINKED ISSUE #97: Setup environment for testing
## 📚 Context
The services running inside the `cartesi-rollups-node` binary depend on external resources: a PostgreSQL database, a Redis instance, and a blockchain provider, to name a few. We should be able to raise an environment containing those resources so we can manually test it during development.

## ✔️ Solution
Adapt the existing `docker-compose.yml` file in the [rollups-examples](https://github.com/cartesi/rollups-examples/blob/main/docker-compose.yml) repo to our needs. Instead of running each service in a container, it'll run a single container with the `cartesi-rollups-node` binary managing all services.

torives:
As `cartesi-rollups-node` gets built and new services are added, it becomes increasingly cumbersome to set up an environment to test it. The added docker-compose file is meant to simplify the process of manual testing for the time being.

Obs: the `server_manager` service will later be removed from the docker-compose to be managed by the `cartesi-rollups-node` binary itself.

Closes #97

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vscode
 .env
 .DS_Store
+build/deployments
 **/.idea/
--- a/build/docker-compose.yml
+++ b/build/docker-compose.yml
@@ -1,0 +1,177 @@
+version: "3.9"
+
+name: rollups-node
+services:
+  rollups-node:
+    build:
+      context: ..
+      dockerfile: build/Dockerfile
+      tags:
+        - "cartesi/rollups-node:devel"
+    entrypoint: ["cartesi-rollups-node", "validator"]
+    ports:
+      - "4000:4000"
+    depends_on:
+      hardhat:
+        condition: service_healthy
+      deployer:
+        condition: service_completed_successfully
+      hardhat_set_interval:
+        condition: service_completed_successfully
+      server_manager:
+        condition: service_healthy
+      database:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
+    environment:
+      RUST_LOG: info
+      CARTESI_LOG_LEVEL: info
+      POSTGRES_ENDPOINT: postgres://postgres:password@database:5432/postgres
+
+      #GraphQL Server
+      GRAPHQL_HEALTHCHECK_PORT: 8082
+      GRAPHQL_HOST: "0.0.0.0"
+      GRAPHQL_PORT: "4000"
+
+      #Indexer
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,172 @@
...
+      dockerfile: build/Dockerfile
+      tags:
+        - "cartesi/rollups-node:devel"
+    entrypoint: ["cartesi-rollups-node", "validator"]
+    ports:
+      - "4000:4000"
+    depends_on:
+      hardhat:
+        condition: service_healthy
+      deployer:
```
marcelstanley, line 16: Just out of curiosity, we're still not at the point where the following setting would be necessary, right?

```
      hardhat_set_interval:
        condition: service_completed_successfully
```
https://github.com/cartesi/rollups-examples/blob/80be07fb68ba7048223c72f3f9fe1b2042ec9ae7/docker-compose.yml#L46-L47

Dockerfile
```diff
@@ -0,0 +1,30 @@
+# (c) Cartesi and individual authors (see AUTHORS)
```
gligneul, line 1: This file should be in the build dir.

docker-compose.yml
```diff
@@ -0,0 +1,172 @@
+version: "3.9"
```
gligneul, line 1: I think this file should be in the build dir as well, but that is up for discussion.

gligneul (commented):
@torives, my understanding is that this was part of https://github.com/cartesi/rollups-node/issues/29. @fmoura

Dockerfile
gligneul: We could rename it to machine.Dockerfile

docker-compose.yml
torives: Not really, as we're still missing the other node services. But to be completely honest, I can't remember exactly the issue `hardhat_set_interval` was meant to solve. Nevertheless, it is a good idea to leave the configuration as close as possible to what we already have, so I'll add the dependency 😉

docker-compose.yml
torives: I see no problem at all 👌🏽

torives:
> @torives, my understanding is that this was part of #29. @fmoura

In the sense that this was a first step, yes. But since #29 is going to take a little while longer to be ready and I needed to advance with #80, I created this based on @fmoura's initial work 😉

gligneul (approved):

marcelstanley (approved):

PR #144, opened 2023-11-09
feat: change config to allow env vars for deployment file data

### LINKED ISSUE #132: Support dapp-deployment-file and rollups-deployment-file config as env vars
We want to add support for environment variables for configuration we currently get from files:

- Authority Claimer
    - --dapp-deployment-file <DAPP_DEPLOYMENT_FILE>
    - --rollups-deployment-file <ROLLUPS_DEPLOYMENT_FILE>

- Dispatcher
    - --rd-dapp-deployment-file <RD_DAPP_DEPLOYMENT_FILE>
    - --rd-rollups-deployment-file <RD_ROLLUPS_DEPLOYMENT_FILE>

------

Data from dapp-deployment-file:
- dapp_address: Address
- dapp_deploy_block_hash: Hash
  
Data from rollups-deployment-file
- history_address: Address
- authority_address: Address
- input_box_address: Address

------

Adding the following environment variables:
- DAPP_ADDRESS
- DAPP_DEPLOY_BLOCK_HASH
- HISTORY_ADDRESS
- AUTHORITY_ADDRESS
- INPUT_BOX_ADDRESS

renan061:


offchain/authority-claimer/src/config/cli.rs
```diff
@@ -10,20 +10,17 @@ use log::{LogConfig, LogEnvCliConfig};
 use rollups_events::{BrokerCLIConfig, BrokerConfig};
```
gligneul, line 10: We should add a change in CHANGELOG mentioning this new behaviour.

offchain/types/src/blockchain_config.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub rollups_deployment_file: PathBuf,
+}
+
+#[derive(Clone, Debug)]
+pub struct BlockchainConfig {
+    pub dapp_address: Address,
+    pub dapp_deploy_block_hash: Hash,
+    pub history_address: Address,
+    pub authority_address: Address,
+    pub input_box_address: Address,
```
gligneul, line 69: So, in the end, we didn't need the deployment hashes of these contracts?
Also, do we need the history address? I think we don't.

offchain/authority-claimer/src/config/cli.rs
gligneul: For instance, `- Added X, Y, and Z env vars to dispatcher and authority claimer. These are optional and will overwrite the values from the rollups deployment file.`

offchain/types/src/blockchain_config.rs
renan061: We only use the hash for the dapp contract. The `Contract` struct is shared.
Going to remove the history contract!

renan061: Don´t think we can remove the history contract.

gligneul (approved):

PR #150, opened 2023-11-14
Fix duplicate checker verification

GMKrieger:
Read the NewClaimToHistory event from the History contract instead of calling the getClaim method from the Authority contract.

### DIFF
--- a/offchain/authority-claimer/src/checker.rs
+++ b/offchain/authority-claimer/src/checker.rs
@@ -2,20 +2,23 @@
 // SPDX-License-Identifier: Apache-2.0 (see LICENSE)
 
 use async_trait::async_trait;
-use contracts::authority::Authority;
+use contracts::history::{Claim, History};
 use ethers::{
     self,
-    abi::AbiEncode,
-    providers::{Http, HttpRateLimitRetryPolicy, Provider, RetryClient},
-    types::{H160, U256},
+    contract::ContractError,
+    providers::{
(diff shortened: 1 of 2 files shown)


gligneul (approved):
Looks good. Let's wait for @renan061 review.

marcelstanley:
LGTM

renan061 (changes requested):
We need to look into the duplicates issue.

offchain/authority-claimer/src/checker.rs
```diff
@@ -83,32 +93,53 @@ impl DuplicateChecker for DefaultDuplicateChecker {
...
-                // bigger than the current one, which means it's a new claim
-                if e.to_string().contains("InvalidClaimIndex()") {
-                    Ok(false)
-                } else {
-                    info!("{:?}", e);
-                    Err(e).context(ContractSnafu)
-                }
-            }
+            .context(ProviderSnafu)?
+            .as_u64();
```
gligneul, line 119: This will not work properly if a reorg happens. With the state-fold code, we looked N blocks behind to be sure. We should do the same here.

gligneul (changes requested):

offchain/authority-claimer/src/checker.rs
GMKrieger: Just pushed a fix that should solve this issue

renan061 (approved):

gligneul (approved):

PR #153, opened 2023-11-16
Add remaining services to `cartesi-rollups-node`

### LINKED ISSUE #139: Add remaining services
## 📚 Context
Now that all the necessary infrastructure to execute the Rust binaries from `cartesi-rollups-node` is in place, all that is left is to start the remaining services (advance-runner, inspect-server, dispatcher, state-server) and test them together to ensure the node's behavior remains the same.

Since the dispatcher depends on the state-server, it'll be necessary to declare such a dependency in code to ensure services start in the correct order.

## ✔️ Solution
Declare the new services and modify `service.Run()` to only start a new service after the previous one returned from a call to `Ready()`, a new method in the service interface that blocks until the service is ready to receive connections or a timeout occurs.

torives:
Closes #139

### DIFF
--- a/build/docker-compose.yml
+++ b/build/docker-compose.yml
@@ -2,7 +2,8 @@ version: "3.9"
 
 name: rollups-node
 services:
-  rollups-node:
+  node:
+    image: "cartesi/rollups-node:devel"
     build:
       context: ..
       dockerfile: build/Dockerfile
@@ -11,6 +12,7 @@ services:
     entrypoint: ["cartesi-rollups-node", "validator"]
     ports:
       - "4000:4000"
+      - "5005:5005"
     depends_on:
       hardhat:
         condition: service_healthy
@@ -28,6 +30,31 @@ services:
       RUST_LOG: info
       CARTESI_LOG_LEVEL: info
       POSTGRES_ENDPOINT: postgres://postgres:password@database:5432/postgres
+      CHAIN_ID: 31337 # dispacher + indexer
+
+      #Advance Runner
+      ADVANCE_RUNNER_HEALTHCHECK_PORT: 8080
+      SERVER_MANAGER_ENDPOINT: http://server_manager:5001
+      PROVIDER_HTTP_ENDPOINT: http://hardhat:8545
+      SESSION_ID: default_rollups_id
+      SNAPSHOT_DIR: /var/opt/cartesi/machine-snapshots
+      SNAPSHOT_LATEST: /var/opt/cartesi/machine-snapshots/latest
+
+      #Authority Claimer
+      AUTHORITY_CLAIMER_HTTP_SERVER_PORT: 8085
+      TX_PROVIDER_HTTP_ENDPOINT: http://hardhat:8545
+      TX_CHAIN_ID: 31337
+      TX_CHAIN_IS_LEGACY: ${TX_LEGACY:-false}
+      TX_DEFAULT_CONFIRMATIONS: 2
+      TX_SIGNING_MNEMONIC: "test test test test test test test test test test test junk"
+
+      #Dispatcher
+      DISPATCHER_HTTP_SERVER_PORT: 8081
+      DAPP_DEPLOYMENT_FILE: /deployments/localhost/dapp.json
+      ROLLUPS_DEPLOYMENT_FILE: /opt/cartesi/share/deployments/localhost.json
+      RD_EPOCH_DURATION: 86400
+      SC_GRPC_ENDPOINT: http://0.0.0.0:50051
+      SC_DEFAULT_CONFIRMATIONS: 1
 
       #GraphQL Server
       GRAPHQL_HEALTHCHECK_PORT: 8082
@@ -38,7 +65,19 @@ services:
       INDEXER_HEALTHCHECK_PORT: 8083
       DAPP_CONTRACT_ADDRESS_FILE: /deployments/localhost/dapp.json
       REDIS_ENDPOINT: redis://redis:6379
-      CHAIN_ID: 31337
+
+      #Inspect Server
+      INSPECT_SERVER_HEALTHCHECK_PORT: 8084
+      INSPECT_SERVER_ADDRESS: 0.0.0.0:5005
+      SERVER_MANAGER_ADDRESS: server_manager:5001
+
+      #State Server
+      SS_SERVER_ADDRESS: 0.0.0.0:50051
+      SF_GENESIS_BLOCK: 0x1
+      SF_SAFETY_MARGIN: 1
+      BH_HTTP_ENDPOINT: http://hardhat:8545
+      BH_WS_ENDPOINT: ws://hardhat:8545
+      BH_BLOCK_TIMEOUT: 8
 
     volumes:
       - machine:/var/opt/cartesi/machine-snapshots
--- a/internal/node/node.go
+++ b/internal/node/node.go
@@ -1,0 +1,90 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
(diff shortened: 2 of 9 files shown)


internal/services/service.go
```diff
@@ -45,33 +46,57 @@ func (s simpleService) Start(ctx context.Context) error {
...
-func Run(services []Service) {
+func Run(ctx context.Context, services []Service) {
 	if len(services) == 0 {
 		logger.Error.Panic("there are no services to run")
 	}
 
 	// start services
-	ctx, cancel := context.WithCancel(context.Background())
-	exit := make(chan struct{})
+	startedServicesCount := 0
```
gligneul, line 97: We can replace part of this logic with a WaitGroup.
https://pkg.go.dev/sync#example-WaitGroup

I will write a PR on top of your proposing something.

internal/services/service.go
```diff
@@ -104,17 +140,76 @@ func Run(services []Service) {
 	case <-wait:
 		logger.Info.Println("main: all services were shutdown")
 	case <-time.After(DefaultServiceTimeout):
-		logger.Warning.Println("main: exited after timeout")
+		logger.Warning.Println("main: exited after a timeout")
+	}
+}
+
+func healthcheckPort(serviceName string) string {
+	if serviceName == "state-server" {
```
gligneul, line 148: Please create a separate function for the state server.
```
func stateServerHealthcheckPort() string {
```

test/fakeservice/main.go
```diff
@@ -0,0 +1,10 @@
+// This file creates a dummy webserver with the sole pupose of being used
```
gligneul, line 1: Put this file inside of ./internal/services/mockservice/

internal/services/service_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/cartesi/rollups-node/internal/logger"
+)
+
+func setup() {
+	logger.Init("debug", false)
+	buildFakeService()
+}
+
+func TestService(t *testing.T) {
```
gligneul, line 23: The tests left a zombie fake-service running on my machine. We need to make sure we are cleaning all resources.

internal/services/service.go
```diff
@@ -104,17 +140,76 @@ func Run(services []Service) {
...
 	}
-	Indexer Service = simpleService{
-		serviceName: "indexer",
-		binaryName:  "cartesi-rollups-indexer",
+	StateServer = Service{
+		name:            "state-server",
+		binaryName:      "cartesi-rollups-state-server",
+		healthcheckPort: healthcheckPort("state-server"),
 	}
 )
```
gligneul, line 215: I feel that this is in the wrong place. This pkg doesn't need to know about the instantiation of the node. Probably, this should be on the cmd or a `node` pkg.

internal/services/service.go
```diff
@@ -1,38 +1,39 @@
...
-	// canceled
-	Start(ctx context.Context) error
-}
-
-const DefaultServiceTimeout = 15 * time.Second
+const (
+	DefaultServiceTimeout  = 15 * time.Second
+	DefaultDialInterval    = 100 * time.Millisecond
+	DefaultStateServerPort = "50051"
+	DefaultHealthcheckPort = "8080"
```
gligneul, line 25: I don't like using a default port because it leads to port conflicts.

gligneul:
See https://github.com/cartesi/rollups-node/pull/162

internal/services/service_test.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Builds the fake-service binary and adds it to PATH
+func buildFakeService() {
+	rootDir, err := filepath.Abs("../../")
+	if err != nil {
+		panic(err)
+	}
+
+	cmd := exec.Command("go", "build", "-o", "build/fake-service", "test/fakeservice/main.go")
```
gligneul, line 136: I don't like the idea of putting this binary in the build dir because this mixes up compilation results with committed files.
I see two options:
1. During the test, create a temp dir, use it as an output dir, and remove this dir at the end of the test.
2. Create a new directory in the project called `bin` or `target` that contains the compilation output.

internal/services/service.go
torives: > Can't we make it explicit every time?

Yes, we can. Also, It's better to keep it all together so it becomes easier to throw away when #155 is merged.

internal/services/service.go
torives: > This pkg doesn't need to know about the instantiation of the node

I agree. I'll put it in `validator.go` for now and after we get rid of Cobra I'll revisit this.

torives: Actually, I changed my mind. There's enough boilerplate code already to justify a `node` package.

internal/services/service_test.go
```diff
@@ -0,0 +1,147 @@
...
+	s.Nil(err, "service timed out")
+
+	// shutdown
+	cancel()
+	s.Nil(<-result, "service exited for the wrong reason: %v", err)
+}
+
+// Builds the fake-service binary and adds it to PATH
+func (s *ServicesTestSuite) buildFakeService() {
+	rootDir, err := filepath.Abs("../../")
```
gligneul, line 120: Why do we need this?

gligneul (changes requested):
Looks good. Minor comments.

internal/node/node.go
```diff
@@ -0,0 +1,89 @@
...
+
+var ValidatorServices = []services.Service{
+	*StateServer,
+	*AdvanceRunner,
+	*AuthorityClaimer,
+	*Dispatcher,
+	*GraphQLServer,
+	*Indexer,
+	*InspectServer,
+}
```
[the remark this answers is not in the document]
torives: Thinking this through, I see no point in using pointers (pun intended). Let's go with values 😉

internal/node/node.go
```diff
@@ -0,0 +1,89 @@
...
+func healthcheckPort(serviceName string) string {
+	env := healthcheckEnv(serviceName)
+	if port, ok := os.LookupEnv(env); ok {
+		if serviceName == "state-server" {
+			split := strings.Split(port, ":")
+			return split[1]
+		}
+		return port
+	} else {
+		return "8080"
```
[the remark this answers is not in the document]
torives: Good point. I'll remove it.

internal/services/service_test.go
```diff
@@ -0,0 +1,147 @@
...
+	logger.Init("warning", false)
+	s.buildFakeService()
+	s.servicePort = 55555
+}
+
+func (s *ServicesTestSuite) TearDownSuite() {
+	err := os.RemoveAll(s.tmpDir)
+	if err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
torives: I hadn't thought about using assert/require for functions that are not proper tests. After trying it I found that using `panic` provides a better description of what happened since it prints the error message instead of the error object itself. 

I would rather keep this as is, but we can discuss this further if you like.

internal/services/service_test.go
torives: Indeed, `rootDir` is not needed anymore. I'll remove it.

gligneul (dismissed):

GMKrieger (dismissed):
LGTM

gligneul (dismissed):

gligneul (approved):

marcelstanley (approved):

build/docker-compose.yml
```diff
@@ -38,7 +65,19 @@ services:
       INDEXER_HEALTHCHECK_PORT: 8083
       DAPP_CONTRACT_ADDRESS_FILE: /deployments/localhost/dapp.json
```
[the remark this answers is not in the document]
gligneul: This will change soon when we have the unified config that Renan is working on.

PR #155, opened 2023-11-16
feat: add go config

renan061:


internal/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+		panic("missing tx-signing environment variables")
+	}
+	return AuthAWS{*keyID, *region}
+}
+
+// ------------------------------------------------------------------------------------------------
+// Custom Types
+// ------------------------------------------------------------------------------------------------
+
+type LogLevel uint8
```
gligneul, line 81: Maybe create a `loglevel.go` file for the LogLevel type and the toLogLevel function.

internal/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+const (
+	LogLevelDebug = iota
+	LogLevelInfo
+	LogLevelWarning
+	LogLevelError
+)
+
+// ------------------------------------------------------------------------------------------------
+
+type Auth interface{}
```
gligneul, line 92: > Maybe create a loglevel.go file for the LogLevel type and the toLogLevel function.

The same applies to the auth-related declarations.

internal/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+		return "", false
+	}
+}
+
+// Parses a string using a given function.
+// It panics on a parsing error, otherwise returns the parsed value.
+func parse[T any](s string, f func(string) (T, error)) T {
+	v, err := f(s)
+	if err != nil {
+		panic(err)
```
gligneul, line 137: I don't like to use panic here because it will explode with an stack trace and will frighten the users. Let's print a nice message and do os.exit.

internal/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	return v
+}
+
+// Returns nil or the value of an environment variable.
+//
+// If the variable could not be read from the environment and it has a default value,
+// then this function will set the cache with the default value and return it parsed.
+func getOptional[T any](name, default_ string, hasDefault bool, parser func(string) (T, error)) *T {
```
gligneul, line 146: I think it is strange to return a pointer in this function. Instead, we should follow the map interface and return the value itself and a boolean.

internal/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	return nil
+}
+
+// Same as getOptional, but panics instead of returning nil.
+func get[T any](name, defaultValue string, hasDefault bool, parser func(string) (T, error)) T {
+	v := getOptional(name, defaultValue, hasDefault, parser)
+	if v == nil {
+		panic(fmt.Errorf(`missing required %s env var`, prefix+name))
```
gligneul, line 167: He should not use panic here as well.
Let's use fmt.Fprintf(stderr, ...) + exit.
Maybe create a function to wrap this behavior.

internal/config/generate/Config.toml
```diff
@@ -0,0 +1,215 @@
+# (c) Cartesi and individual authors (see AUTHORS)
+# SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+# ----------------------------------------
+# Auth -----------------------------------
+# ----------------------------------------
+
+[TX_SIGNING_MNEMONIC]
```
gligneul, line 8: Let's review this list of variables together.

internal/config/generate/Config.toml
```diff
@@ -0,0 +1,215 @@
+# (c) Cartesi and individual authors (see AUTHORS)
+# SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
```
gligneul, line 3: Let's add a variable for the execution mode with an enum.
```
CARTESI_EXECUTION_MODE=validator|reader|no-backend
```

internal/config/generate/Config.toml
```diff
@@ -0,0 +1,215 @@
...
+# ----------------------------------------
+# Auth -----------------------------------
+# ----------------------------------------
```
gligneul, line 6:
```suggestion
#
# Auth
#
```

Even though it looks nice, I'd rather avoid maintaining this ASCII art.

internal/config/mapping.go
```diff
@@ -0,0 +1,37 @@
+// (c) Cartesi and individual authors (see AUTHORS)
```
gligneul, line 1: Let's do this in another issue/PR. We can create this issue together.

internal/config/generate/main.go
```diff
@@ -0,0 +1,133 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package main
```
gligneul, line 4: Please add documentation. 
In particular, about the toml format.

internal/config/config.go
```diff
@@ -0,0 +1,171 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package config
```
gligneul, line 4: Please add documentation.

internal/config/generate/main.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+// function names
+var functions = []string{}
+
+// A value from the toml global level map.
+type EnvGet struct {
+	Default *string `toml:"default"`
+	GoType  string  `toml:"go-type"`
+	Private bool    `toml:"private"`
```
gligneul, line 27: What private means in this case? Is it related to the mapping?

internal/config/get.go
```diff
@@ -0,0 +1,178 @@
...
+func GetBlockchainDefaultConfirmations() int {
+	return get("BLOCKCHAIN_DEFAULT_CONFIRMATIONS", "10", true, toInt)
+}
+
+func GetDappAddress() string { return get("DAPP_ADDRESS", "", false, toString) }
+
+func GetAuthorityClaimerPort() int { return get("AUTHORITY_CLAIMER_PORT", "", false, toInt) }
+
+// Calls all the <Get> functions.
+func getAll() {
```
gligneul, line 138: Shouldn't this be the init function?

internal/config/generate/main.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	// creating the header
+	var builder strings.Builder
+	addLine(&builder, `// (c) Cartesi and individual authors (see AUTHORS)`)
+	addLine(&builder, `// SPDX-License-Identifier: Apache-2.0 (see LICENSE)`)
+
+	addLine(&builder, "")
+	addLine(&builder, `// Code generated by internal/config/generate.`)
+	addLine(&builder, `// DO NOT EDIT.`)
```
gligneul, line 95: If this is the first line of the file my editor opens the file as read-only. Would you change it?

internal/config/generate/main.go
```diff
@@ -0,0 +1,133 @@
...
+	addLine(&builder, `// Calls all the <Get> functions.`)
+	addLine(&builder, `func getAll() {`)
+	sort.Strings(functions)
+	for _, function := range functions {
+		addLine(&builder, `%s()`, function)
+	}
+	addLine(&builder, `}`)
+
+	// assembling the final string
+	s := builder.String()
```
gligneul, line 117: You could simplify the code by writing directly to the out file. I don't see any drawback.

gligneul (commented):
Looks great, just minor comments.

internal/config/get.go
gligneul, line 138: This function won't work as you expect. Depending on the execution mode, the set of required variables will change. And I don't think the config module should be responsible for that.

internal/config/TODO.txt
```diff
@@ -0,0 +1,48 @@
+var mapping = map[string]string{
```
gligneul, line 1: delete this file

internal/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+
+//go:generate go run generate/main.go
+
+// Parsing functions.
+var (
+	toBool     = strconv.ParseBool
+	toInt      = strconv.Atoi
+	toInt64    = func(s string) (int64, error) { return strconv.ParseInt(s, 10, 64) }
+	toString   = func(s string) (string, error) { return s, nil }
+	toDuration = func(s string) (time.Duration, error) { return time.ParseDuration(s + "s") }
```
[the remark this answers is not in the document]
torives: My suggestion is to name it `toSeconds`. Breaks the naming pattern a bit, but it reads better and the return type makes it clear IMO.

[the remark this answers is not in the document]
renan061: I cannot use `toSeconds` because the generated code expects `to<type>`.

I declared the functions separately and used the `var` as aliases.

internal/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+	"time"
+)
+
+//go:generate go run generate/main.go
+
+// Parsing functions.
+var (
+	toBool     = strconv.ParseBool
+	toInt      = strconv.Atoi
+	toInt64    = func(s string) (int64, error) { return strconv.ParseInt(s, 10, 64) }
```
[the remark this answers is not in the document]
renan061: Now using this `var` section for the aliases.

internal/config/config.go
renan061: Learned a neat trick for returning zero values for generic types:
```
var zeroValue T
return zeroValue, false
```

internal/config/generate/Config.toml
renan061: Not a patron of the arts...

 Let's discuss dividers today then. I think it is crucial to be able to divide code visually.

internal/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+		var logLevel = map[string]LogLevel{
+			"debug":   LogLevelDebug,
+			"info":    LogLevelInfo,
+			"warning": LogLevelWarning,
+			"error":   LogLevelError,
+		}
+		if ll, ok := logLevel[s]; ok {
+			return ll, nil
+		} else {
+			return LogLevelDebug, fmt.Errorf(`invalid log level "%s"`, s)
```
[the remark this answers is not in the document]
renan061: Changed it so it uses the zero value.

internal/config/generate/Config.toml
```diff
@@ -0,0 +1,215 @@
...
+
+[BROKER_BACKOFF_MAX_ELAPSED_DURATION]
+go-type = "string"
+
+# ----------------------------------------
+# Postgres -------------------------------
+# ----------------------------------------
+
+[POSTGRES_ENDPOINT]
+default = ""
```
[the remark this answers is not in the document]
renan061: For the Postgres endpoint, the empty string is not an empty value.

Imagine this as `Option<String>`. And this default creates `Some("")`.
Not having the default would create `None`.

internal/config/generate/main.go
renan061: Yes, I will explain it in the docs.

internal/config/config.go
renan061: Created a `types` folder.

internal/config/config.go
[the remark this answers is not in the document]
gligneul: Can't you change the generated code to expect both the name of the function and the type instead?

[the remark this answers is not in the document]
gligneul: Or just the name of the function and have an internal map of the function's name to the type.

internal/config/generate/main.go
renan061: It is easier to use a builder because I format everything using `gofmt`'s inner package string to string directly.

internal/config/types/mode.go
```diff
@@ -0,0 +1,25 @@
...
+	ExecutionModeValidator ExecutionMode = iota
+	ExecutionModeReader
+	ExecutionModeNoBackend
+)
+
+func ToExecutionModeFromString(s string) (ExecutionMode, error) {
+	var m = map[string]ExecutionMode{
+		"validator": ExecutionModeValidator,
+		"reader":    ExecutionModeReader,
+		"no-backed": ExecutionModeNoBackend,
```
torives, line 17:
```suggestion
		"no-backend": ExecutionModeNoBackend,
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added Go config.
```
torives, line 12: If I remember correctly, the consensus was that we don't add modifications to the changelog if the change is not relevant/visible to our users, so I guess we'll add a single entry for the whole go binary, when it becomes ready for production. You can remove this and just add the `no-changelog` tag 😉

internal/config/config.go
```diff
@@ -0,0 +1,171 @@
...
+
+// ------------------------------------------------------------------------------------------------
+// Parsing functions
+// ------------------------------------------------------------------------------------------------
+
+func toInt64FromString(s string) (int64, error) {
+	return strconv.ParseInt(s, 10, 64)
+}
+
+func toStringFromString(s string) (string, error) {
```
gligneul, line 32:
```suggestion
func forwardString(s string) (string, error) {
```
Or something like that.

gligneul (changes requested):
Looks good! Just fix the minor comments left.

CHANGELOG.md
torives: @gligneul now you got me confused. Should or shouldn't we add single container updates to the changelog? 🤔

renan061: If it is in the unreleased section, I see no harm in adding them now.

internal/config/auth.go
```diff
@@ -0,0 +1,19 @@
...
+// Auth objects are used to sign transactions.
+type Auth any
+
+// Allows signing through mnemonics.
+type AuthMnemonic struct {
+	Mnemonic     string
+	AccountIndex int
+}
+
+// Allos signing through AWS services.
```
gligneul, line 15:
```suggestion
// Allows signing through AWS services.
```

gligneul:
Please rebase on top of next.

gligneul (approved):

torives (approved):

PR #158, opened 2023-11-16
fix: make blockchain config more optional

gligneul:


offchain/types/src/blockchain_config.rs
```diff
@@ -97,31 +97,52 @@ impl TryFrom<BlockchainCLIConfig> for BlockchainConfig {
...
+            .map(read::<RollupsDeployment>)
+            .transpose()?;
 
         // try to get the values from the environment values
         // default to the files
-        let dapp_address =
-            env_or_file(cli.dapp_address, dapp.address, "dapp_address")?;
+        let dapp_address = env_or_file(
+            cli.dapp_address,
+            dapp.clone().map(|file| file.address).flatten(),
```
gligneul, line 111: The only way I could find to remove this clone is by using `Option.take`, but it is unsafe.
https://doc.rust-lang.org/std/option/enum.Option.html#method.take

gligneul: I did it, but it just gets worse.

GMKrieger (approved):
LGTM

PR #162, opened 2023-11-18
fixup! test: remove Rust binary dependency

gligneul:


internal/services/service.go
```diff
@@ -34,49 +35,47 @@ type Service struct {
 // Start will execute a binary and wait for its completion or until the context
 // is canceled
 func (s Service) Start(ctx context.Context) error {
-	cmd := exec.Command(s.binaryName)
+	cmd := exec.CommandContext(ctx, s.binaryName)
```
gligneul, line 38: Uses CommandContext to simplify the code a little.

internal/services/service.go
```diff
@@ -34,49 +35,47 @@ type Service struct {
...
 		return err
 	}
+	err := cmd.Run()
+	if err != nil {
+		exitCode := cmd.ProcessState.ExitCode()
+		signal := cmd.ProcessState.Sys().(syscall.WaitStatus).Signal()
+		if exitCode != 0 && signal != syscall.SIGTERM {
+			// only return error if the service exits for reason other than shutdown
+			return err
+		}
```
gligneul, line 56: I changed this to make the code clearer, but the previous version was fine as well.

internal/services/service.go
```diff
@@ -34,49 +35,47 @@ type Service struct {
...
 			return ctx.Err()
-		default:
-			conn, err := net.Dial("tcp", fmt.Sprintf("0.0.0.0:%s", s.healthcheckPort))
-			if err == nil {
-				logger.Debug.Printf("%s is ready\n", s.name)
-				conn.Close()
-				return nil
-			}
-			time.Sleep(DefaultDialInterval)
+		case <-time.After(DefaultDialInterval):
```
gligneul, line 78: Use the select with the time.After, so it stops immediately.

internal/services/service.go
```diff
@@ -34,49 +35,47 @@ type Service struct {
...
 }
 
 // Ready blocks until the service is ready or the context is canceled.
 //
 // A service is considered ready when it is possible to establish a connection
 // to its healthcheck endpoint.
-func (s Service) Ready(ctx context.Context) error {
+func (s Service) Ready(ctx context.Context, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
```
gligneul, line 67: Moved this to the function to make the for loop outside easier to understand.

internal/services/service.go
```diff
@@ -94,48 +93,40 @@ func Run(ctx context.Context, services []Service) {
 	}
 
 	// start services
-	startedServicesCount := 0
 	ctx, cancel := context.WithCancel(ctx)
-	exit := make(chan struct{}, len(services))
+	defer cancel()
+	var wg sync.WaitGroup
```
gligneul, line 98: Using WaitGroup wait for the goroutines.

internal/services/service.go
```diff
@@ -94,48 +93,40 @@ func Run(ctx context.Context, services []Service) {
...
 	for _, service := range services {
 		service := service
+		wg.Add(1)
 		go func() {
+			// cancel the context when one of the services finish
+			defer cancel()
```
gligneul, line 104: Use the context itself to handle cancelation instead of using the exit channel.

test/fakeservice/main.go
```diff
@@ -2,9 +2,13 @@
...
-import "net/http"
+import (
+	"net/http"
+	"os"
+)
 
 func main() {
-	err := http.ListenAndServe(":8090", nil)
+	addr := os.Getenv("SERVICE_ADDRESS")
+	err := http.ListenAndServe(addr, nil)
```
gligneul, line 12: Removing magic numbers

internal/services/service_test.go
```diff
@@ -20,108 +19,90 @@ func setup() {
...
+			name:            "fake-service",
+			binaryName:      "fake-service",
+			healthcheckPort: servicePort,
 		}
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
-		startErr := make(chan error)
-		success := make(chan struct{})
+		// start service in goroutine
```
gligneul, line 39: I simplified all the test cases quite a bit. I think I maintained the same behavior.

torives (approved):
There's nothing like using the APIs properly! It simplifies it quite a bit, thanks for the input. I've learned a thing or two 😉

PR #168, opened 2023-11-22
Add script for generating a test devnet

### LINKED ISSUE #149: Add shell script `gen-devnet` to generate the Anvil state
## 📚 Context

We want to simplify the generation of the Anvil state for the devnet.

## ✔️ Solution

Create a new Go binary under `cmd/gen-devnet` responsible for building the Anvil state. This binary will:
- Receive the following command line parameters:
  - DApp machine template hash (used to deploy the dapp)
  - Path to the anvil state file output
  - Path to the deployment info file (we need a file that contains the addresses of the deployed contracts)
- Download the rollups contracts (see [`pkg/contracts/generate/main.go`](https://github.com/cartesi/rollups-node/blob/next/pkg/contracts/generate/main.go))
- Start Anvil in the background with `--dump-state` (https://github.com/foundry-rs/foundry/tree/master/crates/anvil)
- Deploy the rollups contracts using forge (https://book.getfoundry.sh/forge/deploying)
- Call the authority/history/dapp factory to deploy the authority, history, and DApp contracts using cast (https://book.getfoundry.sh/cast/)
- Stop anvil

We will run this command later inside a Dockerfile. So, it can be assumed that all necessary tools (e.g., foundry) are installed.

marcelstanley:
Closes #149

### DIFF
--- a/scripts/devnet/gen-devnet.sh
+++ b/scripts/devnet/gen-devnet.sh
@@ -1,0 +1,363 @@
+#!/usr/bin/env bash
+# (c) Cartesi and individual authors (see AUTHORS)
+# SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+set -o nounset
+set -o pipefail
+if [[ "${TRACE-0}" == "1" ]]; then set -o xtrace; fi
+
+script_dir="$( cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+readonly script_dir
+
+. "$script_dir/lib/util.sh"
+. "$script_dir/lib/anvil.sh"
+. "$script_dir/lib/contracts.sh"
+
+################################################################################
+# Configuration
+ROLLUPS_CONTRACTS_VERSION="1.1.0"
+DEVNET_RPC_URL="http://localhost:8545"
+DEVNET_FOUNDRY_ACCOUNT_0_ADDRESS="0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
+DEVNET_FOUNDRY_ACCOUNT_0_PRIVATE_KEY="0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
+# keccak256("salt")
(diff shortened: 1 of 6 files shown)


gligneul (commented):
This is starting to look good :)

scripts/devnet/lib/contracts.sh
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env sh
```
marcelstanley, line 1:
```suggestion
#!/usr/bin/env bash
```

scripts/devnet/lib/util.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env sh
```
marcelstanley, line 1:
```suggestion
#!/usr/bin/env bash
```

scripts/devnet/gen-devnet.sh
```diff
@@ -0,0 +1,363 @@
...
+
+   exit 0
+}
+
+################################################################################
+# Finish processing and clean up environment
+finish() {
+    verbose "finishing up"
+
+    if [ -n "$work_dir" ]; then
```
gligneul, line 71: If `check_error` gets called, this will never be executed, leaving the work dir behind.

gligneul: Not a problem for now because this script will be run inside docker build.

gligneul (approved):
This amount of bash makes me uneasy. We should spend little time improving this code because only a hacker like @marcelstanley can maintain it. We should move on, and I will create an issue to rewrite this in Go.

scripts/devnet/gen-devnet.sh
marcelstanley: Not really. This trap should guarantee that `finish` is executed no matter what:

https://github.com/cartesi/rollups-node/pull/168/files#diff-39ace69d7418eeb86b4569890b7a7f56e6a334ee8b235d854c5dc3d4cae7389cR323

fmoura (approved):

PR #171, opened 2023-11-23
ci: add Github Action to check license header

torives:
Adds a new CI workflow to check for the license header in source files.

The action creates warnings for the files not covered in the config file, which is not a big deal but quite annoying. Unfortunately, I couldn't find a way to suppress them, but I'm open to suggestions.

Here are examples of [failed](https://github.com/cartesi/rollups-node/actions/runs/6974420025) and [successful](https://github.com/cartesi/rollups-node/actions/runs/6974497672) runs.


gligneul (dismissed):
Very nice! I'm worried about the maintainability of having various small workflows for small jobs. But I won't block your PR because of that.

torives:
> Very nice! I'm worried about the maintainability of having various small workflows for small jobs. But I won't block your PR because of that.

The changelog and image cleanup workflows have specific triggers, so they need to stay separate. But this and the conventional commit check could run after a push, so I could put them both in the same workflow. What do you say?

marcelstanley:
> The changelog and image cleanup workflows have specific triggers, so they need to stay separate. But this and the conventional commit check could run after a push, so I could put them both in the same workflow. What do you say?

[go-code-quality.yml](https://github.com/cartesi/rollups-node/blob/next/.github/workflows/go-code-quality.yml) seems to be a good candidate for that merge as well.

torives:
> [go-code-quality.yml](https://github.com/cartesi/rollups-node/blob/next/.github/workflows/go-code-quality.yml) seems to be a good candidate for that merge as well.

I thought that wouldn't work, but since the Go linter and the license check steps will ignore files unrelated to their tasks, I guess there's no problem putting them all together. Thanks for the suggestion!

gligneul:
Please rebase this on top of next

gligneul (approved):
Looks good!

.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,34 @@
+name: Assess code quality
```
marcelstanley, line 1: Perhaps we could change the script and the name in the future. I don't have a suggestion right now. 😉

marcelstanley (approved):

PR #175, opened 2023-11-28
feat: start server-manager as an internal service

### LINKED ISSUE #140: Run server-manager as a service
## 📚 Context
Currently, the server-manager is being created as an external dependency inside a docker-compose file. It should be treated as an internal service and therefore also managed by the `cartesi-rollups-node` binary.

Its image (defined in machine.Dockerfile) is also used to create the Cartesi Machine of the DApp we're using in our tests. We still need to create this Cartesi Machine for everything to work.

## ✔️ Solution
Create a new `Service` variable for the server-manager, considering it is available in the PATH when executed inside the `cartesi/rollups-node` docker image like all Rust services and adapt the docker-compose.yml file accordingly.

The Cartesi Machine build process will be moved to its final place in another [issue](https://github.com/cartesi/rollups-node/issues/173).

torives:
Closes #140

### DIFF
--- a/build/docker-compose.yml
+++ b/build/docker-compose.yml
@@ -7,8 +7,6 @@ services:
     build:
       context: ..
       dockerfile: build/Dockerfile
-      tags:
-        - "cartesi/rollups-node:devel"
     entrypoint: ["cartesi-rollups-node", "validator"]
     ports:
       - "4000:4000"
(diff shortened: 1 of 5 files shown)


build/docker-compose.yml
```diff
@@ -164,23 +167,13 @@ services:
         "http://hardhat:8545",
       ]
 
-  server_manager:
+  build_machine:
```
gligneul, line 170: Could we rename this to `machine_snapshot_setup`?

gligneul (changes requested):
Just one minor comment; other than that, it looks good.

GMKrieger (approved):

gligneul (approved):

PR #176, opened 2023-11-29
Fixes for dispatcher and claimer

### LINKED ISSUE #166: Dispatcher finish epoch at wrong time after getting state from broker
## 🙂 Expected behavior

The dispatcher should finish the rollups epoch in a deterministic manner.
So, if we restart the dispatcher and it loads the current state from the broker, it should finish the following epochs as if it wasn't restarted in the first place.

## 🫠 Actual behavior

The dispatcher finishes the epoch at the wrong time after it restarts.
See the test case for more details.

@endersonmaia detected this bug in a staging environment with Sepolia testnet.
The DApp address used by the node that triggered the bug is `0x0000000000000000000000004ca2f6935200b9a782a78f408f640f17b29809d8`.
Here is the event log from the History contract: https://sepolia.etherscan.io/address/0x76f4dcac0920826541ee718214eee4be07346cee#events

By looking at the contract event log, we can see that the node generated the claim with inputs number 6-7 [(claim 3)](https://sepolia.etherscan.io/tx/0x338f84cdc02cf415412ccdf8f8cb3740f6f0f03463bcb729516916858905188b) a few minutes after generating the claim with input number 5 ([claim 2](https://sepolia.etherscan.io/tx/0xf66b17d6157881a838eca5474920a5035701d23c29bb70fa3ce18099c2ff75f8)).

By looking at the dispatcher logs provided by Enderson, I noted that the dispatcher restarted just before making claim number 2.

## 🧪 Minimal test case

I reproduced the issue in examples with the following script:

1. Start the dispatcher
2. Send input
3. Advance EVM time by epoch duration
4. Wait for the dispatcher to finish the first epoch
5. Restart the dispatcher
6. Send two inputs
7. The dispatcher finishes the second epoch wrongly (with just the first input, instead of two of them).

## 🌎 Environment

Rollups Node 1.1.

## ✔️ Possible solutions

When starting up, the dispatcher sets the timestamp for the latest event as the genesis timestamp, regardless of the rollups state in the broker.
https://github.com/cartesi/rollups-node/blob/main/offchain/dispatcher/src/drivers/context.rs#L39

### LINKED ISSUE #169: Node fails send submit claim if it doesn't exists
## 🙂 Expected behavior

The node should submit the following claim expected by the contracts.

## 🫠 Actual behavior

No node tries to submit a claim that doesn't exist but is not expected by the contracts.

## 🧪 Minimal test case

This might happen if a previous run of the dispatcher finishes the epoch too soon.
Then, rerunning the dispatcher will generate a different set of claims.
This issue is described in https://github.com/cartesi/rollups-node/issues/166.

## 🌎 Environment

Rollups 1.0 and 1.1.

gligneul:
Close #166
Close #169


offchain/dispatcher/src/setup.rs
```diff
@@ -94,15 +94,21 @@ pub async fn create_context(
         .timestamp
         .as_u64();
     let epoch_length = config.epoch_duration;
+
+    let status = broker.status().await.context(BrokerSnafu)?;
+
+    // The dispatcher doesn't work properly if there are inputs in the broker from a previous run.
+    // Hence, we make sure that the broker is in a clean state before starting.
```
gligneul, line 101: I gave up trying to solve the bug because we should always restart Redis when the node restarts. Hence, the dispatcher will fail during startup when there is an input in Redis. This will be improved when we rewrite the node in Go.

torives (dismissed):

GMKrieger (dismissed):

offchain/dispatcher/src/drivers/context.rs
```diff
@@ -27,21 +27,19 @@ impl Context {
     pub async fn new(
```
GCdePaula, line 27: I think `async` can be removed.

offchain/dispatcher/src/setup.rs
GCdePaula: :/ what a bummer.

State all spread around is very annoying.

GCdePaula (approved):
LGTM!

torives (dismissed):

torives (approved):

PR #184, opened 2023-12-04
OpenAPI for application management

tuler:
Defines an openapi for application management of a (future) multi-application node.
It defines 4 endpoints: listing, adding, querying and removing.

Listing is paginated and has filtering based on status.

I used `application` instead of `dapp`. We're favoring `application` in several locations including sunodo codebase and rollups-explorer codebase, because there are several tools that generate names based on that, and the term "DApp", or "dApp" does not behave well with those tools, often generating oddities like `d-app`, or `d_app`.


api/openapi.yaml
```diff
@@ -0,0 +1,209 @@
...
+          type: string
+          format: url
+        status:
+          type: string
+          description: application status
+          enum:
+            - started
+            - starting
+            - stopping
+            - stopped
```
gligneul, line 188: When do these two cases happen? Will we keep the applications as stopped once we delete them?

api/openapi.yaml
```diff
@@ -0,0 +1,209 @@
...
+            - error
+  requestBodies:
+    Application:
+      description: Application to be added to the node
+      required: true
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
```
gligneul, line 198: We should probably also add:
- epoch duration
- whether the app will run in reader mode

gligneul (changes requested):
Looking good

api/openapi.yaml
```diff
@@ -0,0 +1,209 @@
...
+paths:
+  /applications:
+    get:
+      tags:
+        - applications
+      summary: Fetches list of applications managed by the node
+      description: List of managed applications
+      operationId: getApplications
+      parameters:
+        - name: status
```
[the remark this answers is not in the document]
tuler: I left intentionally as non-required. But I don't know, I think it can be required. I'll change.

api/openapi.yaml
tuler: I can remove the "stopped", but leave the "stopping".

I think there will be the case where a client will issue a DELETE, and poll for the completion of the operation (as DELETE returns 202). Then it will polling waiting for a 404 of the get. It's kind of weird.

api/openapi.yaml
tuler: I don't think the epoch duration whould be part of the API, for simplicity. The node would be configured (offchain) and use the same duration for all application, as this is already an offchain decision. The other params (address, block, etc) are all values that will come from the blockchain.

tuler: How would you define the mode? An enum with values "reader", "validator" and "full". Or a "reader" boolean and a "validator" boolean.

api/openapi.yaml
```diff
@@ -0,0 +1,209 @@
...
+          description: Application address
+          required: true
+          schema:
+            type: string
+            format: address
+            example: "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
+            pattern: "^0x([0-9a-fA-F]{40})$"
+      requestBody:
+        $ref: "#/components/requestBodies/Application"
+      responses:
```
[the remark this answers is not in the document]
tuler: I'll add the 409, but I think the implementation should only return 409 if the params are in conflict with the running dapp. If they are the same we could return 304.

api/openapi.yaml
```diff
@@ -0,0 +1,204 @@
+openapi: 3.0.0
```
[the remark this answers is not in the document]
tuler: Maybe the best path is `/openapi/management.yaml`

endersonmaia:
I found no reference to the chain the node is running?

Is it expected to a multi-dapp node to work on multiple chains, of for a single chain?

Anyway, I think some API endpoint (or response header) should expose details for the configuration of the node, the chainID being one of these details.

```http
GET /v1/status
{ "chainID":  "123", "epochDuration": "68000", "applicationCount": 10, ...}
 ```

**Comment by tuler** [comment]
> Is it expected to a multi-dapp node to work on multiple chains, of for a single chain?

No. It's single chain.

> Anyway, I think some API endpoint (or response header) should expose details for the configuration of the node, the chainID being one of these details.

The process running the API will receive the same env variables as a node would receive.
Aiming for simplicity, I don't think we need to expose this to an endpoint.

**Comment by tuler** [comment]
> "applicationCount": 10

Application count is part of the API in the response of the `/applications`

File: api/openapi.yaml (hunk 1 above)
Comment (gligneul) [reply api/openapi.yaml]: Yes, this is weird. We can leave the stopped to avoid that.

File: api/openapi.yaml (hunk 2 above)
Comment (gligneul) [reply api/openapi.yaml]: We won't provide a validator mode without a reader mode in Monodo because everything will be under the same database. That said, we currently use a boolean to indicate whether the node will run in reader mode. We could be consistent here. https://github.com/cartesi/rollups-node/blob/de4b8bd4d6fa86e9c6bf3e1436cafc06825ad941/internal/config/generate/Config.toml#L24

File: api/management.openapi.yaml [hunk 6]
```diff
@@ -0,0 +1,243 @@
...
+          schema:
+            $ref: "#/components/schemas/Address"
+      responses:
+        "200":
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Application"
+        "400":
```
gligneul, line 80: Could you reuse common responses? See https://swagger.io/docs/specification/describing-responses/#reuse

api/openapi.yaml
gligneul: Getting the block number might be a pain. See https://levelup.gitconnected.com/how-to-get-smart-contract-creation-block-number-7f22f8952be0

Maybe viem can come to the rescue.

api/management.openapi.yaml
```diff
@@ -0,0 +1,243 @@
...
+      parameters:
+        - name: address
+          in: path
+          description: Application address to delete
+          required: true
+          schema:
+            $ref: "#/components/schemas/Address"
+      responses:
+        "202":
+          description: Delete request accepted
```
endersonmaia, line 143: This response could have a `Link: ` header with the status of the request that could be polled.

tuler: But there would be no link to refer to. What would it be?

api/management.openapi.yaml
tuler: Done

api/management.openapi.yaml
```diff
@@ -0,0 +1,238 @@
...
+      requestBody:
+        $ref: "#/components/requestBodies/Application"
+      responses:
+        "200":
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Application"
+        "304":
```
tuler, line 106: I think I'll remove this 304. It's not common to return 304 on a PUT request. It's usually used on a GET with `If-Modified-Since`.
I think the correct way is to just return 200 if the resource is exactly the same.

api/openapi.yaml
tuler: The API will be called upon a blockchain event, the block number won't be a problem.

api/management.openapi.yaml
endersonmaia: - /delete-requests/123
- /jobs/123

gligneul (dismissed):
Looks good!

api/management.openapi.yaml
tuler: It becomes too complicated

tuler:
Rebased, squashed, renamed file.

gligneul (dismissed):

gligneul:
>  I'm working on the implementation, and it looks like the case will be that we will send a SIGTERM to gracefully shutdown a node, and wait for it to die with some timeout. If it succeeds we can return 204, if it times out we can return a 202. Does it make sense?

So, will it be a sync request? Do you plan to remove the `stopping` and `stopped` status?

tuler:
> So, will it be a sync request? Do you plan to remove the `stopping` and `stopped` status?

It will be a sync request. If times out the client would have to query the GET until it returns `stopped`, or if we remove the `stopped` until it returns 404.

gligneul:
> It will be a sync request. If times out the client would have to query the GET until it returns stopped, or if we remove the stopped until it returns 404.

I see. I don't have strong opinions about the stopped, you can leave it if you want.

tuler:
`stopped` status removed

vfusco:
The management API will be a json-rpc api. It will be based on what the node cli can do, that is more than what was described here. Closing so this don't cause confusion. It's still referred on the issue #255

PR #203, opened 2023-12-11
feat: add configuration documentation

### LINKED ISSUE #177: Add `description` tag to Config.toml
## 📈 Subtasks

- [x] Code `generate.go` to document the functions.
- [x] Integrate with `gendocs` command (`.md` file).
- [x] Add the `description` text in the Config.toml file.
- [x]  Log new environment variables when getting them

renan061:


cmd/cartesi-rollups-cli/cartesi-rollups-cli
gligneul, line 1: Remove this binary.

gligneul (changes requested):

cmd/gen-docs/gen-docs
gligneul, line 1: Please remove this binary.

internal/config/config.md
gligneul, line 1: This file should be in `/docs`.

internal/config/generate/main.go
```diff
@@ -17,24 +17,197 @@ import (
...
+
+	return topicNames, topicEnvs
+}
+
+func addLine(builder *strings.Builder, s string, a ...any) {
+	builder.WriteString(fmt.Sprintf(s, a...))
+	builder.WriteString("\n")
+}
+
+func addCodeHeader(builder *strings.Builder) {
```
gligneul, line 99: Could you move functions related to the printing to two separate files? One for code printing and another one for docs printing.

gligneul (changes requested):

internal/config/config_test.go
torives, line 1: It's not a big deal, but I'd argue that using the [suite](https://pkg.go.dev/github.com/stretchr/testify/suite) API for the tests would improve their readability. Setup/teardown code gets extracted and the tests within code blocks can become individual testing functions. It takes a bit more work, but IMO it is worth it for the sake of clarity. 

We're already using it in other test files, so it would be nice to maintain the same style (for non-trivial unit tests at least).

renan061: The tests that require a setup are sequential, so we cannot split them using suite. :-[

internal/config/config.go
```diff
@@ -19,7 +19,7 @@ import (
 	"time"
 )
 
-//go:generate go run generate/main.go
+//go:generate go run generate/helpers.go generate/main.go
```
gligneul, line 22:
```suggestion
//go:generate go run ./generate
```

internal/config/config_test.go
```diff
@@ -0,0 +1,260 @@
...
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRead(t *testing.T) {
+	require := require.New(t)
```
gligneul, line 18: I think we don't need this line. Could you check?

internal/config/config_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+	cacheLen := 2
+	require.Equal(cacheLen, len(cache.values)) // from the init() of log.go
+	envSetup()                                 // clears the cache
```
gligneul, line 22: It would be best to use the suite pkg to avoid manually calling this setup. You could use the same setup for all the tests.

internal/config/config_test.go
```diff
@@ -0,0 +1,260 @@
...
+func TestToLogLevelError(t *testing.T) {
+	v, err := toLogLevel("error")
+	require.Nil(t, err)
+	require.Equal(t, LogLevelError, v)
+}
+
+func TestToLogLevelFail(t *testing.T) {
+	_, err := toLogLevel("not log level")
+	require.NotNil(t, err)
+}
```
gligneul, line 171: These tests could be in another file or another section.

gligneul (commented):
Looks good just minor comments.

internal/config/config_test.go
renan061: This allows us to call the require functions, like `require.Equal`, without having to pass the `t *testing.T` parameter all the time.

gligneul (approved):

torives (approved):

PR #210, opened 2023-12-12
fix: service reachability

torives:
When executing the cartesi-rollups-node in a docker container, using the address 127.0.0.1 blocks services from receiving connections from interfaces outside the container. This prevents users from interacting with the Host Runner or querying the GraphQL and Inspect servers.

This commit also fixes the port mapping for GraphQL and Inspect servers in the node-compose.yml file

---
**Obs:** I've only changed the IP addresses of the services that need to receive connections from outside the container. I'm unsure of what is best: to make the difference apparent or to standardize it all.

@GMKrieger thank you for the help in understanding what was going on.

### DIFF
--- a/build/node-compose.yml
+++ b/build/node-compose.yml
@@ -5,8 +5,8 @@ services:
   node:
     image: "cartesi/rollups-node:devel"
     ports:
-      - "10003:4000"
-      - "10007:5005"
+      - "10004:10004" # GraphQL Server
+      - "10009:10009" # Inspect Server
     restart: always
     depends_on:
       devnet:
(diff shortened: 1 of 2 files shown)


endersonmaia (commented):
I think the default in `services.go` should be 127.0.0.1, and you could define an environment variable at `node-compose.yml` that would override the default to work inside the container.

gligneul (commented):
This is going to change soon when we implement the proxy. When we have the proxy, we should expose only its port; the other addresses could be localhost.

gligneul (approved):

GMKrieger (approved):

PR #212, opened 2023-12-13
Add external healthcheck to `cartesi-rollups-node`

### LINKED ISSUE #193: Setup health check endpoint

torives:
Closes #193

### DIFF
--- a/cmd/cartesi-rollups-node/services.go
+++ b/cmd/cartesi-rollups-node/services.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 
 	"github.com/cartesi/rollups-node/internal/config"
@@ -325,4 +326,18 @@ func newSupervisorService(s []services.Service) services.SupervisorService {
 		Name:     "rollups-node",
 		Services: s,
 	}
+}
+
+func newHttpService() services.HttpService {
+	handler := http.NewServeMux()
+	handler.Handle("/healthz", http.HandlerFunc(healthcheckHandler))
+	return services.HttpService{
+		Name:    "http",
+		Address: fmt.Sprintf("%v:%v", config.GetCartesiHttpAddress(), getPort(portOffsetProxy)),
+		Handler: handler,
+	}
+}
+
+func healthcheckHandler(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(http.StatusOK)
 }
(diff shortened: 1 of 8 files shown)


cmd/cartesi-rollups-node/services.go
```diff
@@ -310,3 +311,10 @@ func newStateServer() services.Service {
 	s.Env = append(s.Env, os.Environ()...)
 	return s
 }
+
+func newHealthService() services.Service {
```
gligneul, line 315: The health check shouldn't be a service on its own. It should be part of an HTTP service that will have multiple endpoints. One of them is the health check. The proxy defines the other endpoints. For inspiration, see https://github.com/gligneul/nonodo/blob/main/internal/nonodo/echo.go.

internal/services/health.go
```diff
@@ -0,0 +1,66 @@
...
+		return err
+	case <-ctx.Done():
+		if err := server.Close(); err != nil {
+			logger.Warning.Printf("%v: error while closing: %v\n", s.String(), err)
+		}
+		return ctx.Err()
+	}
+}
+
+func (s HealthService) Ready(ctx context.Context, timeout time.Duration) error {
```
gligneul, line 46: This abstraction doesn't translate well in this case. See https://github.com/gligneul/nonodo/blob/main/internal/supervisor/supervisor.go#L25 and https://github.com/gligneul/nonodo/blob/main/internal/nonodo/echo.go#L54.

For the binary service, see: https://github.com/gligneul/nonodo/blob/main/internal/supervisor/command.go#L52

gligneul (changes requested):

internal/services/http.go
```diff
@@ -0,0 +1,54 @@
+// (c) Cartesi and individual authors (see AUTHORS)
```
gligneul, line 1: Let's add a test for this file.

cmd/cartesi-rollups-node/services.go
```diff
@@ -326,3 +327,17 @@ func newSupervisorService(s []services.Service) services.SupervisorService {
 		Services: s,
 	}
 }
+
+func newHttpService() services.HttpService {
+	handler := http.NewServeMux()
+	handler.Handle("/healthz", http.HandlerFunc(healthcheckHandler))
+	return services.HttpService{
+		Name:    "http",
+		Address: fmt.Sprintf("127.0.0.1:%v", getPort(portOffsetProxy)),
```
gligneul, line 336:
```suggestion
		Address: fmt.Sprintf("%v:%v", config.GetCartesiHttpAddress(), getPort(portOffsetProxy)),
```

gligneul: We need to add this variable, the default should be 127.0.0.1

gligneul (approved):

PR #220, opened 2023-12-19
fix: panic when executable not in $PATH

### LINKED ISSUE #205: Go Node segfaults when binary is not found
## 🙂 Expected behavior

Raise a readable error

## 🫠 Actual behavior

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x4e0660]

goroutine 34 [running]:
os.(*ProcessState).sys(...)
	/usr/local/go/src/os/exec_posix.go:91
os.(*ProcessState).Sys(...)
	/usr/local/go/src/os/exec.go:170
github.com/cartesi/rollups-node/internal/services.Service.Start({{0x51e257, 0xe}, 0x2afd, {0x52267e, 0x1e}, {0x0, 0x0, 0x0}, {0xc000142000, 0x4c, ...}}, ...)
	/home/gligneul/cartesi/rollups-node/internal/services/service.go:70 +0x220
github.com/cartesi/rollups-node/internal/services.Run.func1()
	/home/gligneul/cartesi/rollups-node/internal/services/service.go:124 +0xb9
created by github.com/cartesi/rollups-node/internal/services.Run in goroutine 1
	/home/gligneul/cartesi/rollups-node/internal/services/service.go:120 +0x305
exit status 2
```

## 🧪 Minimal test case

Start the Go node without one of the binaries in the PATH.

## 🌎 Environment

main

torives:
The Service struct used to evaluate its termination reason to decide if it would return an error or not. This code didn't check for nil pointers and could panic during runtime.

Instead of evaluating the exit code or retrieving the received signal, we changed the design to always exit with an error. It now can be a `context.Canceled` or `exec.ExitError`.

The new code will ignore errors occurring during the exit process after context cancelation. This is by design since this is a rare edge case and the Supervisor is only interested in terminating the process.

Closes #205

### DIFF
--- a/internal/services/service_test.go
+++ b/internal/services/service_test.go
@@ -60,7 +60,7 @@ func (s *ServiceTestSuite) TestServiceStops() {
 	// shutdown
 	cancel()
 	err := <-result
-	s.Nil(err, "service exited for the wrong reason: %v", err)
+	s.ErrorIs(err, context.Canceled, "service exited for the wrong reason: %v", err)
 }
 
 // Service should stop if timeout is reached and it isn't ready yet
@@ -85,7 +85,8 @@ func (s *ServiceTestSuite) TestServiceTimeout() {
 
 	// shutdown
 	cancel()
-	s.Nil(<-result, "service exited for the wrong reason: %v", err)
+	err = <-result
+	s.ErrorIs(err, context.Canceled, "service exited for the wrong reason: %v", err)
 }
 
 // Service should be ready soon after starting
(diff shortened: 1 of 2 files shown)


internal/services/service_test.go
```diff
@@ -110,7 +111,24 @@ func (s *ServiceTestSuite) TestServiceReady() {
...
+		Name:            "fake-service",
+		Path:            "wrong-path",
+		HealthcheckPort: s.servicePort,
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	s.NotPanics(func() {
+		err := service.Start(ctx)
+		s.NotNil(err)
```
gligneul, line 130: could you check the error?

internal/services/service_test.go
```diff
@@ -110,7 +111,24 @@ func (s *ServiceTestSuite) TestServiceReady() {
 
 	// shutdown
 	cancel()
-	s.Nil(<-result, "service exited for the wrong reason: %v", err)
+	err = <-result
+	s.ErrorIs(err, context.Canceled, "service exited for the wrong reason: %v", err)
+}
+
+// Regression test to catch a panic if the executable is not found in $PATH
```
renan061, line 118: Minor of minors, but: we don't need to document this as a "Regression test".
Just `test to catch a panic if the executable is not found in $PATH` is fine.

renan061 (dismissed):

gligneul (dismissed):

renan061 (approved):

gligneul (approved):

PR #228, opened 2024-01-05
Refactor/docker build enhancements

endersonmaia:


build/Dockerfile
torives, line 1: I guess we should also modify lines 43 and 44 to use RUN <<EOF

torives (commented):
Thanks for the improvements 😉

gligneul:
@endersonmaia could you squash?

torives (approved):

Issue #183, opened 2023-12-04
Verify whether generated files are in sync


gligneul:
## 📚 Context

We are committing generated text files to the repo because it is more straightforward. However, if someone forgets to update the generated file, it will get out of sync.

I would rather not commit these files in CI or generate them on demand.

## ✔️ Solution

Add a CI job that generates the files and checks if they are out-of-sync. If that is the case, the person will be forced to fix their PR before merging it.

Currently, we generate files with:
- `go generate`
- `go run ./cmd/gen-docs`
- This file https://github.com/cartesi/rollups-node/blob/main/offchain/graphql-server/src/schema/generate_schema.rs

## 🐢 

Also, let's add an "auto-generated" warning to the files as discussed at https://github.com/cartesi/rollups-node/pull/170#discussion_r1413803041

marcelstanley:
Also, let's add a "auto-generated" warning to the files as discussed at https://github.com/cartesi/rollups-node/pull/170#discussion_r1413803041

gligneul:
We should probably set-up a `tools.go` file to track the Go tools we use with `go generate`. See https://marcofranssen.nl/manage-go-tools-via-go-modules

PR #231, opened 2024-01-08
Check if generated files are in sync

GMKrieger:
#183

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+.PHONY: all generate cargo diff
+
+all: diff
+
+generate:
+	@echo "Generating files in tools.go"
+	@go generate -v ./...
+
+cargo:
+	@cd offchain; cargo run --bin generate-schema
+	@mv offchain/schema.graphql api/graphql/reader.graphql
+
+diff: cargo generate
+	@echo "Checking differences on the repository..."
+	@if git diff --exit-code; then \
+		echo "No differences found."; \
+	else \
+		echo "ERROR: Differences found in the resulting files."; \
+		exit 1; \
+	fi
--- a/README.md
+++ b/README.md
@@ -24,6 +24,14 @@ Before building and running any of the inner projects, you should download the s
 git submodule update --init --recursive
 ```
 
+### Generating files
+
+All the necessary auto-generated files are already commited to the project; but if you wish to generate them again, you can run the Makefile at the root of the project. Simply run the command:
+
+```shell
+make
+```
+
 ### Building the Docker image
 
 For more information on how to build the rollups-node docker image, see the [build directory](./build/README.md).
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -32,3 +32,6 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         with:
           version: v1.54.2
+
+      - name: Check auto generated files
+        run: make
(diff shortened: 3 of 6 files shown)


tools.go
```diff
@@ -0,0 +1,13 @@
...
+
+//go:build ignore
+
+package main
+
+import (
+	_ "cmd/gen-docs"
+	_ "internal/config/config.go"
+	_ "pkg/contracts/main.go"
+	_ "pkg/readerclient/main.go"
```
gligneul, line 12: I don't get the purpose of this file as is. As far as I know, the tools file is supposed to include only the **external** Go binary tools so they are included in the mods file.

Makefile
```diff
@@ -0,0 +1,20 @@
+.PHONY: all generate cargo diff
+
+all: diff
+
+generate:
+	@echo "Generating files in tools.go"
+	@cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go generate %
```
gligneul, line 7: Why not `go generate -v ./...`?

Makefile
```diff
@@ -0,0 +1,20 @@
+.PHONY: all generate cargo diff
```
gligneul, line 1: We need some documentation about this file in the README.

Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+generate:
+	@echo "Generating files in tools.go"
+	@cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go generate %
+
+cargo: generate
+	@cd offchain; cargo run --bin generate-schema
+	@mv offchain/schema.graphql api/graphql/reader.graphql
+
+diff: cargo
```
gligneul, line 13:
```suggestion
cargo: 
	@cd offchain; cargo run --bin generate-schema
	@mv offchain/schema.graphql api/graphql/reader.graphql

diff: cargo generate
```

It makes more sense to put generate as a dependency of diff instead of cargo.

gligneul (changes requested):

tools.go
GMKrieger: Hm, I was not aware of this distinction. I used it as a compiler so every go generate could be easily looked up in a single file. Looking it now, I realize I just patched on a solution for another problem online on the code and, since it worked, I carried on. Fixing it now.

Makefile
GMKrieger: Same thing as with `tools.go`

README.md
```diff
@@ -24,6 +24,14 @@ Before building and running any of the inner projects, you should download the s
 git submodule update --init --recursive
 ```
 
+### Generating files
+
+You can run the Makefile in the projet to generate the project files auto-generated files, simply run the command at the root:
```
gligneul, line 29: Please explain we commit the generated files and this step is only required when changing them.

Makefile
```diff
@@ -0,0 +1,23 @@
+.PHONY: all generate cargo diff
+
+all: diff
+
+generate:
+	@echo "Generating files in tools.go"
+	@go generate -v ./cmd/gen-docs
+	@go generate -v ./internal/config/config.go
+	@go generate -v ./pkg/contracts/main.go
+	@go generate -v ./pkg/readerclient/main.go
```
gligneul, line 10: We should use `go generate ./...` instead of explicitly declaring each file here.

gligneul (changes requested):

gligneul:
@GMKrieger, is anything else missing from the PR? If not, could you make it "ready for review"?

gligneul (approved):

PR #240, opened 2024-01-11
Create SupervisorService

### LINKED ISSUE #221: Create SupervisorService
## 📚 Context

As a supervisor, the `cartesi-rollups-node` main role is to start and manage different services. We would like to thoroughly test such important functionality but this effort is currently hindered by its current implementation as we have no easy way of knowing all services have started.

## ✔️ Solution

Use the `Service` abstraction to create a service that executes other services. This would bring the added benefit of opening the opportunity for new types of services to be created as well.

## 📈 Subtasks

- [x] Extract a Service interface from `Service` struct
- [x] Rename the `Service` struct to `CommandService`
- [x] Use a channel to communicate readiness instead of the `Ready()` function
- [ ] Abstract current `services.Run()` logic inside a new `SupervisorService`

### LINKED ISSUE #222: Implement graceful shutdown
## 📚 Context

The supervisor should exit gracefully when receiving a SIGINT or a SIGTERM, stopping its services and releasing their resources appropriately.

## ✔️ Solution

Create the main supervisor context with [signal.NotifyContext](https://pkg.go.dev/os/signal#NotifyContext)

torives:
Closes #221, closes #222

### DIFF
--- a/internal/services/command_test.go
+++ b/internal/services/command_test.go
@@ -39,8 +39,7 @@ func (s *CommandServiceSuite) SetupTest() {
 	os.Setenv("SERVICE_ADDRESS", serviceAdress)
 }
 
-// Service should stop when context is cancelled
-func (s *CommandServiceSuite) TestItStops() {
+func (s *CommandServiceSuite) TestItStopsWhenContextIsCancelled() {
 	service := CommandService{
 		Name:            "fake-service",
 		Path:            "fake-service",
@@ -60,11 +59,11 @@ func (s *CommandServiceSuite) TestItStops() {
 	select {
 	case err := <-result:
 		s.FailNow("service failed to start", err)
-	case <-time.After(100 * time.Millisecond):
+	case <-ready:
 	}
 
-	// shutdown
 	cancel()
+
 	err := <-result
 	s.ErrorIs(err, context.Canceled, "service exited for the wrong reason: %v", err)
 }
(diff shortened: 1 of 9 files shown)


internal/services/command_test.go
```diff
@@ -51,11 +51,17 @@ func (s *ServiceTestSuite) TestServiceStops() {
...
-		result <- service.Start(ctx)
+		result <- service.Start(ctx, ready)
 	}()
 
-	time.Sleep(100 * time.Millisecond)
+	// assert service started successfully
+	select {
+	case err := <-result:
+		s.FailNow("service failed to start", err)
+	case <-time.After(100 * time.Millisecond):
```
gligneul, line 63:
``` suggestion
	case <-ready:
```

internal/services/command_test.go
```diff
@@ -74,24 +80,23 @@ func (s *ServiceTestSuite) TestServiceTimeout() {
...
 	}()
 
 	// expect timeout because of wrong port
-	err := service.Ready(ctx, 500*time.Millisecond)
-	s.NotNil(err, "expected service to timeout")
-
-	// shutdown
-	cancel()
-	err = <-result
-	s.ErrorIs(err, context.Canceled, "service exited for the wrong reason: %v", err)
```
gligneul, line 89: This should not be removed.

internal/services/command_test.go
```diff
@@ -74,24 +80,23 @@ func (s *ServiceTestSuite) TestServiceTimeout() {
...
+	case <-ready:
+		s.FailNow("service should have timed out")
+	case <-time.After(2 * time.Second):
+	}
 }
 
 // Service should be ready soon after starting
-func (s *ServiceTestSuite) TestServiceReady() {
-	service := Service{
+func (s *CommandServiceSuite) TestItBecomesReady() {
```
gligneul, line 98: This test case is redundant with `TestItStops` and should be removed.

gligneul (changes requested):

gligneul (approved):
nice

Issue #217, opened 2023-12-18
Add command to run dependencies on docker


fmoura:
This command needs to start all the node dependencies:

- Postgres
- Anvil

We will be using Docker to run them. To help these services being used by automated integration tests, testcontainers package should be used

PR #249, opened 2024-01-16
build: use compose build configuration

endersonmaia:
This PR has no much impact, but I think it's more natural when using `docker compose` to run `build` and then `up`.

But that's me! :)


gligneul (commented):
Most likely we will stop using docker-compose to run the node. See 

- https://github.com/cartesi/rollups-node/issues/217
- https://github.com/cartesi/rollups-node/issues/29

gligneul:
Replaced by #265

PR #260, opened 2024-01-17
refactor: Docker build system

gligneul:
- Refactored the Dockerfile, adding comprehensive documentation.
- Moved the version config to the bake file.
- Added the rollups-node prefix to all Docker images.
- Changed the installation dir of the binaries to /usr/bin.
- Changed the base Docker image to debian-bookworm instead cartesi/server-manager.


build/Dockerfile
```diff
@@ -174,24 +224,30 @@ apt-get install -y --no-install-recommends \
...
+# Setup runtime dir.
+ARG RUNTIME_DIR=/usr/share/rollups-node
 WORKDIR ${RUNTIME_DIR}
+RUN chown cartesi:cartesi ${RUNTIME_DIR}
 
+# Set user back to low-privilege.
 USER cartesi
-CMD ["cartesi-rollups-node"]
+
+# Set the Go supervisor as the entrypoint.
```
endersonmaia, line 252: I suggest not using the naming entrypoint here to not confuse with Dockerfile's `ENTRYPOINT` instrunction.

pkg/ethutil/ethutil_test.go
```diff
@@ -93,10 +93,7 @@ func TestEthUtilSuite(t *testing.T) {
 // This image starts an anvil node with the Rollups contracts already deployed.
 func newDevNetContainer(ctx context.Context) (testcontainers.Container, error) {
 	req := testcontainers.ContainerRequest{
-		Image: "cartesi/rollups-devnet:devel",
-		Env: map[string]string{
-			"ANVIL_IP_ADDR": "0.0.0.0",
```
endersonmaia, line 98: Removing the `ANVIL_IP_ADDR: 0.0.0.0` may cause problems to connect to it from other container.

I can't tell if this environment is being defined elsewhere, just ignore my comment if that's the case.

build/Dockerfile
```diff
@@ -3,131 +3,171 @@
 
 # syntax=docker.io/docker/dockerfile:1.4
```
endersonmaia, line 4:
```suggestion
# syntax=docker.io/docker/dockerfile:1
```

I suggest using only major version `1`, since it guarantees we'll have access to new functionalities and bugfixes from the builder.

https://docs.docker.com/build/dockerfile/frontend/#stable-channel

build/Dockerfile
```diff
@@ -138,32 +178,40 @@ apt-get install -y --no-install-recommends \
...
 COPY . .
-RUN go install -ldflags "-s -w" ./cmd/cartesi-rollups-node
+RUN go build -ldflags "-s -w" ./cmd/cartesi-rollups-node
 
+# STAGE: rollups-node
 #
-# Runtime
-#
+# This stage copies the binaries from the previous stages.
 FROM cartesi/server-manager:${SERVER_MANAGER_VERSION} as rollups-node
```
endersonmaia, line 209: I'm uncomfortable with this; maybe we could make some changes now.

We should have more control over the runtime image, having all we need explicit in this file.

When we use `FROM server-manager` we get another `FROM machine-emulator` and it's difficult to track what's inside those images.

My suggestion is to use `FROM debian:bookworm-slim` (or even use the sha256 hash for reproducibility and security) and copy server-manager binaries and install the machine-emulator .deb package.

Something like this should be enough, and check for library dependencies.

NOT TESTED CODE
```
FROM debian:bookworm-slim
...
COPY --from=cartesi/server-manager:${SERVER_MANAGER_VERSION} /usr/bin/server-manager /usr/bin/server-manager
...
RUN <<EOF
wget https://github.com/cartesi/machine-emulator/releases/download/$MACHINE_EMULATOR_VERSION/cartesi-machine-v$MACHINE_EMULATOR_VERSION_$ARCH.deb
apt install -y cartesi-machine-v$MACHINE_EMULATOR_VERSION_$ARCH.deb
rm cartesi-machine-v$MACHINE_EMULATOR_VERSION_$ARCH.deb
EOF
....
RUN <<EOF
addgroup --system --gid 102 adduser --system --uid 102 --ingroup cartesi --disabled-login --no-create-home --home /nonexistent --gecos "cartesi user" --shell /bin/false cartesi
```

endersonmaia (changes requested):

build/Dockerfile
```diff
@@ -3,131 +3,171 @@
...
-CMD ["--block-time", "5", "--load-state" , "/usr/share/devnet/anvil_state.json"]
+# Start Anvil.
+CMD anvil --block-time 5 --load-state $ANVIL_STATE_PATH
 
+####################################################################################################
+# TARGET: rollups-node
 #
-# Cargo chef
+# This target is the final Docker image that will be used in the production environment.
+# It install all the binaries necessary to run the node in /usr/bin.
```
torives, line 144:
```suggestion
# It installs all the binaries necessary to run the node in /usr/bin.
```

build/Dockerfile
```diff
@@ -138,32 +178,40 @@ apt-get install -y --no-install-recommends \
...
-COPY --from=rust-planner ${RUST_BUILD_PATH}/recipe.json .
+# Build external dependencies with cargo chef.
+COPY --from=rust-prepare ${RUST_BUILD_PATH}/recipe.json .
 RUN cargo chef cook --release --recipe-path recipe.json
 
-# Build application
+# Build application.
 COPY ./offchain/ .
-# needed to log commit hash in runtime
-COPY ./.git ../
```
torives, line 148: I know you did this intentionally, but just for the record: this removes commit info from Rust services' log, but speeds up image-building time by preventing cache invalidation.

torives (commented):
It looks great, it became way easier to reason about the file 👍🏽

build/Dockerfile
endersonmaia: I'll send a PR over to this branch for your consideration.

endersonmaia: https://github.com/cartesi/rollups-node/pull/263

gligneul: Thanks. I'll use this PR as inspiration to create a separate stage to be the base image. This image should be used when building the Cartesi machine for the `rollups-node-snapshot` target.

build/Dockerfile
```diff
@@ -1,133 +1,217 @@
...
+# The result is used as the base for the snapshot and the node targets.
+# We do this instead of using the cartesi/machine-emulator image to have control over the distro
+# used by the base image.
+FROM ${BASE_IMAGE} as emulator-base
+
+# Install machine-emulator
+ARG MACHINE_EMULATOR_VERSION
+RUN <<EOF
+apt-get update
+apt-get install -y curl
```
endersonmaia, line 44:
```suggestion
apt-get install -y --no-install-recommends
    ca-certificates \
    curl
```

We save 4 packages, but I won´t block for this. :)

build/Dockerfile
```diff
@@ -172,28 +266,38 @@ apt-get install -y --no-install-recommends \
...
+ARG RUNTIME_DIR=/usr/share/rollups-node
 WORKDIR ${RUNTIME_DIR}
+RUN chown cartesi:cartesi ${RUNTIME_DIR}
 
+# Set user to low-privilege.
 USER cartesi
-CMD ["cartesi-rollups-node"]
+
+# Set the Go supervisor as the command.
+CMD cartesi-rollups-node
```
endersonmaia, line 303: This way `PID 1` will be sinde `sh -c`.

I'm assuming the go-supervisor is traping the signals, so it's better to use the exec form of `CMD`

```suggestion
CMD [ "cartesi-rollups-node" ]
```

See: https://docs.docker.com/engine/reference/builder/#shell-and-exec-form

endersonmaia (changes requested):

endersonmaia (approved):
LGTM :rocket:

PR #263, opened 2024-01-18
Feature/distroless container images

endersonmaia:


gligneul:
Added this directly to the base PR.

Issue #259, opened 2024-01-16
Check machine hash in Go


gligneul:
## 📚 Context

When the advance-runner starts, it calls the blockchain app contract to obtain the machine hash and compare it with the loaded snapshot.
I will remove this feature from the advance-runner as part of a simplification; we still need to do this check.

## ✔️ Solution

Do this check on Go during the supervisor's start-up.

Be sure to add integration tests that use the rollups contracts.

PR #267, opened 2024-01-19
refactor: removed snapshot storage

### LINKED ISSUE #196: Advance runner fails to start in later epochs after input stream is wiped
## 🙂 Expected behavior

The advance runner should be resilient and wait for the input stream to be filled instead of failing.

## 🫠 Actual behavior

The advance runner fails to start, raising an error. See https://github.com/cartesi/rollups-node/blob/main/offchain/advance-runner/src/runner.rs#L132

## 🧪 Minimal test case

This is not trivial to reproduce. Here is a suggestion:

1. Start the single node
2. Send inputs
3. Finish epoch
4. Restart the single node
5. The advance runner should fail. (It might not fail if the dispatcher starts first and has time to fill the inputs stream).

## 🌎 Environment

Rollups 1.2.

## ✔️ Possible solutions

Add a backoff + retry mechanism.

gligneul:
We have a few reasons to remove this feature:
1. Currently, it is not being used because the production environment doesn't persist the snapshot.
2. Not having this feature will make the transition to Go easier.
3. It fixes the #196 bug.
4. This feature will not work with the output unification changes in the server manager.

Close #196


offchain/advance-runner/src/dapp_contract.rs
gligneul, line 1: This will be reimplement in Go: https://github.com/cartesi/rollups-node/issues/259

build/Dockerfile
```diff
@@ -20,13 +20,14 @@ ARG LINUX_KERNEL_VERSION
...
-ARG MACHINE_SNAPSHOT_BUILD_PATH=/build/machine-snapshot
+ARG SNAPSHOT_BUILD_PATH=/build/snapshot
 ARG DEVNET_BUILD_PATH=/build/devnet
 ARG RUST_BUILD_PATH=/build/rollups-node/rust
 ARG GO_BUILD_PATH=/build/rollups-node/go
 
 # Runtime dir for the cartesi-machine snapshot.
-ARG MACHINE_SNAPSHOT_RUNTIME_PATH=/usr/share/cartesi/snapshots
+ARG SNAPSHOT_RUNTIME_PATH=/usr/share/cartesi/snapshots
+ARG SNAPSHOT_FILENAME=test-application
```
endersonmaia, line 30: `_FILENAME` is weird since it's a directory.

If the `SNAPSHOT_RUNTIME_PATH` is gonna hold only a single snapshot, maybe renaming it to `/usr/share/cartesi/snapshot` and don't use `SNAPTHOT_FILENAME`.

Issue #192, opened 2023-12-05
Add command to validate notice


gligneul:
## 📚 Context

Validating notices and executing vouchers are central steps in our rollups solution.
We need a command to verify that these steps are working as expected.
We can start with notices because they do not have a specific format to be executed; instead, they just need valid proof to be validated.

## ✔️ Solution

- [ ] Add auxiliary function to `ethutil` package.
- [ ] Add command to validate notice.

PR #270, opened 2024-01-22
Add validate notice command

GMKrieger:
#192

### DIFF
--- a/cmd/cartesi-rollups-cli/root/validate/validate.go
+++ b/cmd/cartesi-rollups-cli/root/validate/validate.go
@@ -1,0 +1,92 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package validate
+
+import (
+	"os"
+
+	"github.com/Khan/genqlient/graphql"
+	"github.com/cartesi/rollups-node/internal/config"
+	"github.com/cartesi/rollups-node/pkg/addresses"
+	"github.com/cartesi/rollups-node/pkg/ethutil"
+	"github.com/cartesi/rollups-node/pkg/readerclient"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/spf13/cobra"
+)
+
+var Cmd = &cobra.Command{
+	Use:     "validate",
+	Short:   "Validates a notice",
+	Example: examples,
+	Run:     run,
+}
+
+const examples = `# Validates notice 5 from input 6:
+cartesi-rollups-cli validate --notice-index 5 --input-index 6`
+
+var (
+	noticeIndex     int
+	inputIndex      int
+	graphqlEndpoint string
+	ethEndpoint     string
+	addressBookFile string
+)
+
+func init() {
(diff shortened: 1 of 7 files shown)


cmd/cartesi-rollups-cli/root/validate/validate.go
```diff
@@ -0,0 +1,94 @@
...
+
+func run(cmd *cobra.Command, args []string) {
+	ctx := cmd.Context()
+	graphqlClient := graphql.NewClient(graphqlEndpoint, nil)
+
+	resp, err := readerclient.GetNotice(ctx, graphqlClient, noticeIndex, inputIndex)
+	cobra.CheckErr(err)
+
+	if resp.Proof == nil {
+		fmt.Print("The notice has no associated proof yet.\n")
```
gligneul, line 64: The function should stop at this point.

gligneul, line 64: Please use `config.InfoLog` instead of `fmt.Print`.

cmd/cartesi-rollups-cli/root/validate/validate.go
```diff
@@ -0,0 +1,94 @@
...
+		inputIndex,
+		book.CartesiDApp,
+	)
+	valid, err := ethutil.CallValidateNotice(client, book, resp.Payload, proof)
+	cobra.CheckErr(err)
+
+	if valid {
+		fmt.Print("The notice is valid!\n")
+	} else {
+		fmt.Print("The notice is not valid...\n")
```
gligneul, line 92: This doesn't get executed. See:

```
➜  rollups-node-2 git:(b4d86fc2) go run ./cmd/cartesi-rollups-cli validate --input-index 0 --notice-index 0
connected to http://localhost:8545
validating notice 0 from input 0 with address 180763470853caf642df79a908f9282c61692a45
Error: execution reverted
exit status 1
```

I changed the proof in the DB with dbeaver to make it fail.

It seems like the contract always return true: https://github.com/cartesi/rollups-contracts/blob/v1.2.0/onchain/rollups/contracts/dapp/CartesiDApp.sol#L186

pkg/readerclient/proof.go
```diff
@@ -96,8 +97,14 @@ func newProof(
 			return nil, fmt.Errorf("failed to decode MachineStateHash to bytes: %v", err)
```
gligneul, line 97: I noticed this error message is wrong. Can you fix it and the others?

gligneul (changes requested):
Looks good! Just minor comments.

pkg/ethutil/ethutil.go
```diff
@@ -107,3 +107,25 @@ func GetInputFromInputBox(
...
+	notice []byte,
+	proof *contracts.Proof,
+) (bool, error) {
+
+	dapp, err := contracts.NewCartesiDApp(book.CartesiDApp, client)
+	if err != nil {
+		return false, fmt.Errorf("failed to connect to CartesiDapp contract: %v", err)
+	}
+
+	response, err := dapp.ValidateNotice(nil, notice, *proof)
```
gligneul, line 125: You should pass the contex.Context here.

gligneul (changes requested):

PR #276, opened 2024-01-23
fix: remove concurrent block fetch in inputbox

GCdePaula:


gligneul:
Looks good! Could you fix the CI?

PR #280, opened 2024-01-24
chore: add set -e for RUN heredoc instruction

endersonmaia:
When replacing RUN apt update && apt install ... we were making use of &&, that ensures that the following command will only execute if the previous one succeeds. When using the RUN <<EOF format, we lose this and can have some surprises  uring the docker build process. It's common for a command inside an RUN heredoc type to have a filing process but that whole instruction pass as a success in the build, and the result will only be noticed on the runtime image.

Adding set -e will ensure that if any command fails, the whole instruction will fail.

See: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -40,6 +40,7 @@ FROM ${BASE_IMAGE} as emulator-base
 # Install machine-emulator
 ARG MACHINE_EMULATOR_VERSION
 RUN <<EOF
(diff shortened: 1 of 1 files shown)


gligneul (approved):

gligneul:
Nice catch

PR #282, opened 2024-01-24
refactor: remove dead code

gligneul:
The dispatcher doesn't use the history foldable because this code was reimplemented in the authority-claimer service.


GCdePaula:
This works, but you can also just change the "root" foldable from `RollupState` to `InputBox`. (Then, if you want, you could delete `RollupsState` and `History`.)

The `RollupsState` is just this aggregate foldable. I recommend not using it, and instead setting the `InputBox` as your foldable.

How can I help you? I can do it if you prefer.

offchain/dispatcher/src/dispatcher.rs
```diff
@@ -57,19 +58,15 @@ pub async fn start(
...
-            .into_inner()
-            .into(),
+    let initial_state = InputBoxInitialState {
+        input_box_address: Arc::new(
+            config
+                .blockchain_config
+                .input_box_address
+                .clone()
+                .into_inner()
+                .into(),
```
GCdePaula, line 68: These two lines looked strange to me, but I think I get it now. There are two definitions for `Address`: one from `ethers.rs` and another from our broker. I imagine to make the rest of the node not depend on `ethers.rs`.

GCdePaula (approved):
LGTM!

PR #289, opened 2024-01-29
Create cartesi-rollups devnet in Go

### LINKED ISSUE #229: Deploy Rollups contracts deterministically on devnet in Go
## 📚 Context

Rollups contracts are not begin deployed in a deterministic fashion on `devnet`, which forced us to change the address book used by `cartesi-rollups-cli` to make the relevant addresses match.

We must amend the deployment process so the whole deployment is deterministic.

## ✔️ Solution

~Use a tool (possibly [forge script](https://book.getfoundry.sh/reference/forge/forge-script)) that uses the `CREATE2` opcode to perform the deployment (see [tutorial](https://book.getfoundry.sh/tutorials/create2-tutorial)).~

~Note: Besides being used by rollups-contracts, there's also a mention about using [safe-singleton-factory](https://github.com/safe-global/safe-singleton-factory) at https://github.com/foundry-rs/foundry/issues/2638#issuecomment-1860269617 pointing to https://github.com/bgd-labs/aave-helpers/blob/6cc788d1346e6bcaf8f02878841eb74662c87c17/src/ScriptUtils.sol#L87.~

Use rollups-contracts deployment tools.

## 📈 Subtasks

- [x] Deploy contracts deterministically on `devnet`
- [x] Restore original address book used by `cartesi-rollups-cli`
- [x] Convert shell code to Go

marcelstanley:
Tasks:
- [x] Deploy contracts deterministically
- [x] Download and deploy rollups-contracts in Go
- [x] Start anvil in Go
- [x] Create Application in Go
- [x] Replace heatlhcheck script with inline code
- [x] Clean up/streamline error handling

New
- [x] add rollups-contracts as submodule
- [x] use internal Service

Closes #229


cmd/gen-devnet/main.go
```diff
@@ -0,0 +1,162 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	rollupsContractsPath, err := filepath.Abs(rollupsContractsPath)
+	if err != nil {
+		config.ErrorLogger.Printf("%v does not exist", err)
+		return
+	}
+
+	// Add machine hash to config
```
marcelstanley, line 89:
```suggestion
```

build/Dockerfile
```diff
@@ -121,58 +121,85 @@ CMD /bin/bash
...
 
+# Install dependencies needed to install yarn
+RUN <<EOF
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+    gnupg
+EOF
+
+# Install yarn
+RUN <<EOF
```
endersonmaia, line 157:
```suggestion
RUN <<EOF
set -e
```

build/Dockerfile
```diff
@@ -121,58 +121,85 @@ CMD /bin/bash
 # STAGE: devnet-base
 #
 # This stage installs Foundry.
-FROM ${BASE_IMAGE} as devnet-base
+FROM golang:${GO_VERSION}-bookworm as devnet-base
 
 # Install system dependencies.
-ARG DEBIAN_FRONTEND=noninteractive
 RUN <<EOF
```
endersonmaia, line 127: To keep consistency in this Dockerfile, use this `ARG` before each RUN that has an `apt` command.

```suggestion
ARG DEBIAN_FRONTEND=noninteractive
RUN <<EOF
```

build/Dockerfile
```diff
@@ -121,58 +121,85 @@ CMD /bin/bash
...
 set -e
 apt-get update
-apt-get install -y --no-install-recommends \
+DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
```
endersonmaia, line 130: If use ARG, remove this here.

```suggestion
apt-get install -y --no-install-recommends \
```

build/Dockerfile
```diff
@@ -121,58 +121,85 @@ CMD /bin/bash
...
 #
 # This stage builds the devnet state that will be loaded in Anvil.
 FROM devnet-base as devnet-builder
 ARG DEVNET_BUILD_PATH
 WORKDIR ${DEVNET_BUILD_PATH}
 
+# Install dependencies needed to install yarn
+RUN <<EOF
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
```
endersonmaia, line 152: same here

build/Dockerfile
```diff
@@ -121,58 +121,85 @@ CMD /bin/bash
...
+    gnupg
+EOF
+
+# Install yarn
+RUN <<EOF
+curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
```
endersonmaia, line 161: same here

build/Dockerfile
```diff
@@ -121,58 +121,85 @@ CMD /bin/bash
...
+
 # STAGE: devnet-builder
 #
 # This stage builds the devnet state that will be loaded in Anvil.
 FROM devnet-base as devnet-builder
 ARG DEVNET_BUILD_PATH
 WORKDIR ${DEVNET_BUILD_PATH}
 
+# Install dependencies needed to install yarn
+RUN <<EOF
```
endersonmaia, line 150: To avoid hiding errors during build.

```suggestion
RUN <<EOF
set -e
```

build/Dockerfile
marcelstanley: Thanks for  the suggestion but, as we talked off line, I'll make sure this setting is uniform throughout the file.

That said, I'll resolve the other comments to keep this suggestion in a single thread.

marcelstanley: Please take a look if the changes made at a18f82f6 suffice.

endersonmaia (approved):

build/Dockerfile
```diff
@@ -121,7 +123,7 @@ CMD /bin/bash
 # STAGE: devnet-base
 #
 # This stage installs Foundry.
-FROM ${BASE_IMAGE} as devnet-base
+FROM golang:${GO_VERSION}-bookworm as devnet-base
```
gligneul, line 126: I think it is strange that the devnet-base, which will be used as the base for the final image, has to Go.
This should be installed in the devnet-builder image only.
Or use the Go in the base for the devnet-builder, and install the foundry twice, one time in the builder and another in the devnet image.

build/Dockerfile
```diff
@@ -131,52 +133,85 @@ apt-get update
...
+curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+apt-get update
+apt-get install -y --no-install-recommends \
+    nodejs \
+    yarn
+EOF
+
+# Copy scripts
+COPY . .
```
gligneul, line 174: We should avoid copying everything because it will invalidate the docker cache. Especially things like the .git directory.

cmd/gen-devnet/deployer.go
```diff
@@ -0,0 +1,130 @@
...
+	CONTRACT_OWNER_ADDRESS = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
+	SALT                   = "0x0000000000000000000000000000000000000000000000000000000000000000"
+)
+
+func deploy(ctx context.Context,
+	rollupsContractsPath string,
+	hash string) (DeploymentInfo, error) {
+	var depInfo DeploymentInfo
+
+	config.InfoLogger.Printf("deployer: deploying %v", rollupsContractsPath)
```
gligneul, line 29: @torives is changing the CLI command not to use the logger but instead just fmt.Printf. These commands don't need a full-blown logger, so I agreed with him in favor of making it simpler. Let's do the same here.

gligneul (dismissed):
Looks great, just minor comments.

build/Dockerfile
marcelstanley: >Or use the Go in the base for the devnet-builder, and install the foundry twice, one time in the builder and another in the devnet image.

I thought about using this option but eventually decided for using the golang image as base but I agree we can do it as you said.

marcelstanley: Yes, it is a bit strange. I really though about doing what you suggested and installing foundry twice. However, installing Go requires downloading the right binaries for the target platform, which seemed a b it convoluted at first. That's why I decided to go the way it is. But, anyway, let me see how it goes this other way.

build/Dockerfile
marcelstanley: It looks like we're doing the same with the node build at https://github.com/cartesi/rollups-node/blob/4c74729249cebd4b0c74e4d4bafd103f96d7eaef/build/Dockerfile#L245.

So, I'll:

- adjust `.gitignore` to not copy any git/github related file; and
- restrict the copied directories to the ones needed to build devnet

endersonmaia: Don't confuse `.gitignore` with `.dockerignore`.

build/Dockerfile
marcelstanley: Changes made at ee968dc7

build/Dockerfile
marcelstanley: Changes made at af52782a and ee968dc7

cmd/gen-devnet/deployer.go
marcelstanley: Changes made at ee968dc7

cmd/gen-devnet/rollups.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if err := deployContracts(ctx, tmpDir); err != nil {
+		return fmt.Errorf("failed to deploy contracts: %v", err)
+	}
+	return nil
+}
+
+// Copy only the rollups hardhat project
+func copyContracts(srcDir string, destDir string) error {
+	return cp.Copy(srcDir, destDir+string(os.PathSeparator)+"rollups")
```
torives, line 50: I'm just curious as to why you chose an external dependency over calling `cp` with `exec.Command`. Was it for portability reasons? Because as of now our code isn't exactly portable 😅

torives (commented):
I'd also recommend adding `anvil_state.json` and `deployment.json` to the gitignore, in case the command is run locally. 

I couldn't read `deployment.json` because of a permission error, It would be interesting to understand what is going on.

cmd/gen-devnet/rollups.go
marcelstanley: Yeah, that's rather unnecessary! Thanks for pointing that out. Changed at b90cc13d

build/Dockerfile
```diff
@@ -131,16 +134,18 @@ apt-get update
...
 # Install Foundry from downloaded pre-compiled binaries
 ARG FOUNDRY_NIGHTLY_VERSION
 RUN curl -sSL https://github.com/foundry-rs/foundry/releases/download/nightly-${FOUNDRY_NIGHTLY_VERSION}/foundry_nightly_linux_$(dpkg --print-architecture).tar.gz | \
-    tar -zx -C /usr/local/bin
+        tar -zx -C /usr/local/bin
+
+# STAGE: devnet-golang
+#
+# This stage makes GO_VERSION available locally
+FROM golang:${GO_VERSION}-bookworm as devnet-golang
```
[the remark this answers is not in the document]
marcelstanley: My bad. Changed at b90cc13d

build/Dockerfile
```diff
@@ -149,34 +154,89 @@ FROM devnet-base as devnet-builder
...
+
+# Copy gen-devnet script and dependencies
+# This should be simplified in the future when `COPY --parents` is made available
+RUN mkdir -p internal/config \
+    && mkdir -p internal/services \
+    && mkdir -p pkg/addresses \
+    && mkdir -p pkg/contracts \
+    && mkdir -p pkg/ethutil \
+    && mkdir -p rollups-contracts \
+    && mkdir -p cmd/gen-devnet
```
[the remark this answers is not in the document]
marcelstanley: Nope. Changed at b90cc13d

build/Dockerfile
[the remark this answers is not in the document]
endersonmaia: > Is there any advantage in creating this extra stage to simply installing Golang directly on the devnet-builder stage?

One reason is the registry's cache. It may be faster to build many times using `FROM ...` than without it.

marcelstanley:
> I'd also recommend adding `anvil_state.json` and `deployment.json` to the gitignore, in case the command is run locally.
> 
> I couldn't read `deployment.json` because of a permission error, It would be interesting to understand what is going on.

Both changes were also made at [c08ab3a](https://github.com/cartesi/rollups-node/pull/289/commits/c08ab3a197c03491ef70481acb1ccd0c0d54416a)

build/Dockerfile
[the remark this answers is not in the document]
marcelstanley: Sure, good point, but in that case it was just a matter of working around an issue and, considering this is an internal tool, I'd rather go with @torives's suggestion and simply install Go on it's own layer.

marcelstanley:
@torives, I just had to perform a push-force due after removing some dead code from `build/Dockerfile`

torives (approved):

Issue #304, opened 2024-02-02
Rename env vars after upgrading to rollups-contracts 2.0.0


torives:
## 📚 Context

There will be several changes in nomenclature in the new major version of the rollup contracts, so it is important to keep our related environment variables in sync with the source.

## ✔️ Solution

Update the variables in `Config.toml` to reflect the changes

## 📈 Subtasks

- [ ] Remove `CARTESI_CONTRACTS_HISTORY_ADDRESS`
- [ ] Rename `CARTESI_CONTRACTS_AUTHORITY_ADDRESS` to `CARTESI_CONTRACTS_CONSENSUS_ADDRESS`

torives:
@guidanoli thanks for bringing this to our attention 😉

guidanoli:
Just a side note that this is not really related to the output unification. It was just a simplification on the consensus part of the contracts. :-)

marcelstanley:
Done as part of #377 at 5dcffde

guidanoli:
Why was this closed as not planned?

torives:
> Why was this closed as not planned?

@guidanoli I think Marcel reopened the issue by accident and just closed it again. The actual code is already merged on the `next/2.0` branch.

PR #305, opened 2024-02-02
refactor: rename env vars

torives:
Renames environment variables to reflect the changes in rollups contracts v2.0.0, where the word "application" is being used instead of "dapp". 

We are still using version 1.2.0 of the contracts, but since we're already introducing breaking changes to the env vars in version 1.3.0 of the Node, I've decided to include these two already. Related to #304.


docs/config.md
gligneul, line 1: I'm reviewing the documentation, and there is another change that makes sense to me.
Could you rename `CARTESI_FEATURE_READER_MODE` to `CARTESI_FEATURE_DISABLE_CLAIMER`.

Rationale:
This week, there was some confusion regarding what the reader mode is. Some people expected the reader mode to be an infinitely scalable solution to run inspects. However, the reader mode just disables the claimer part; the rest of the node works the same. So, renaming this variable will make our intent more explicit.

torives:
@endersonmaia FYI

gligneul (approved):

docs/config.md
endersonmaia: Iw as going to suggest `CARTESI_FEATURE_CLAIMER_ENABLED: true` to maintain in sync with `CARTESI_EXPERIMENTAL_SUNODO_VALIDATOR_ENABLED`.

Like, every boolean config variable should have something like an `_ENABLED` suffix.

So the last one missing I can see is `CARTESI_FEATURE_HOST_MODE`, which could be `CARTESI_FEATURE_HOST_MODE_ENABLED: false`.

PR #318, opened 2024-02-09
feat: Improve config generation

### LINKED ISSUE #309: Improve Node Services' configuration generation
## 📚 Context

The Node configuration generation as of today generates specific public methods for each variable, and these methods get them from System Environment. It makes it difficult to configure the services

## ✔️ Solution

Instead of generating specific methods, generate a Configuration Structure with all the variables. This way the Tests are able to construct the same structure with custom values.

fmoura:
closes #309


torives (commented):
This is looking great! It'll help me log the node's configs easily now in #319.

Since I had the time, I've taken the liberty to propose a few changes in #323. I suggest checking the commit messages for extra context.

After the PR is merged, I'd argue we would need a follow-up issue to refactor `config` package into something more cohesive. The cache feature is mostly useless as it is now, for example. But that's a discussion for later 😉

docs/config.md
```diff
@@ -143,6 +143,7 @@ When enabled, the node does not start the authority-claimer service and the Redi
 External Redis endpoint for the node when running in the experimental sunodo validator mode.
 
 * **Type:** `string`
+* **Default:** `""`
```
renan061, line 146: Why an empty default?

internal/config/generate/Config.toml
```diff
@@ -234,6 +234,7 @@ description = """
 When enabled, the node does not start the authority-claimer service and the Redis server."""
 
 [experimental.CARTESI_EXPERIMENTAL_SUNODO_VALIDATOR_REDIS_ENDPOINT]
+default = ""
```
renan061, line 237: Why an empty default?

internal/config/generate/main.go
```diff
@@ -27,20 +27,86 @@ func main() {
...
+func writeCode(envs []Env) {
 	var code strings.Builder
-	var doc strings.Builder
 
 	addCodeHeader(&code)
-	addDocHeader(&doc)
 
-	addLine(&doc, "<!-- markdownlint-disable MD012 -->")
+	//Validate envs
+	for i := range envs {
```
renan061, line 43: `for _, env := range envs {`

internal/config/generate/Config.toml
torives: It's a workaround/compromise. 

Now that there's a 1:1 map from env vars to NodeConfig member variables, `getCartesiExperimentalSunodoValidatorRedisEndpoint()` is called even if `CARTESI_EXPERIMENTAL_SUNODO_VALIDATOR_ENABLED=false` and we'll get an error if there's no value.

I couldn't find a solution that didn't involve a lot more code, so I went this not-so-elegant route. This is one of the reasons I think another issue is needed to better adapt the `config` package API.

renan061 (changes requested):
It seems strange to me that `NodeConfig` gets initialized from the environment with all its fields set. Are there no optional fields? I think this has to do with my prior comment on the RedisEndpoint.

Before, the program would explode if we required an env that was not set and had no default. But that was ok because we only called Get when the program had to explode if an env was not set and had no default.

But now we always call all gets. Won't that fail in production?

I think some fields should be `*type` and the constructor for `NodeConfig` should call
`getOptional`. Then we can have a (not generated) function `Validate()` that checks whether we have a valid node config instance.

Make sense?

renan061:
> After the PR is merged, I'd argue we would need a follow-up issue to refactor `config` package into something more cohesive. The cache feature is mostly useless as it is now, for example. But that's a discussion for later 😉

I agree that we should refactor `config`. Not sure if we should some of that here (like i proposed).

internal/config/generate/main.go
fmoura: as Validate sets the default value, if we use `_,env` we get a copy of the "Env" and the value is not correctly set. Unless we do `.validate()` on every loop and use this Env instance to generate that code snippet.

internal/config/generate/Config.toml
fmoura: I will take a better look on this now. These "logic" about when to use or how to use the configs are spread out at some functions outside the Config Package.

fmoura:
After some discussions we are going to merge this PR as is . The problems mentioned by @renan061 will be resolved right after on top of the issue #325

renan061:
> After some discussions we are going to merge this PR as is . The problems mentioned by @renan061 will be resolved right after on top of the issue #325

Since this PR will break the main branch, I think it is best to merge the other config PRs on top of this branch.
And then we continue with the proposed changes.

internal/config/custom_get.go
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+// The config package handles all of the node's configurations,
+// which are mostly obtained through environment variables.
+//
+// The get.go file contains most of the getter functions for the configuration values.
+// Its code is automatically generated by the generate/main.go script,
+// given the environment variable configured in the generate/Config.toml file.
+//
+// Custom getters are defined in the config.go file.
```
torives, line 11: This was probably copy-pasted from the `config.go` file and should be deleted.

torives: I also suggest renaming this file to `custom_getters.go`.

internal/config/config_test.go
```diff
@@ -77,57 +77,57 @@ func (suite *EnvSuite) TestRead() {
...
-	require.Equal(suite.T(), foo, v)
+	require.Equal(suite.T(), foo, *v)
 }
 
-func (suite *EnvSuite) TestLogInit() {
-	require.Equal(suite.T(), 2, len(cache.values))
-}
+// func (suite *EnvSuite) TestLogInit() {
+// 	require.Equal(suite.T(), 2, len(cache.values))
+// }
```
torives, line 130: This can be deleted

internal/config/config.go
```diff
@@ -48,37 +48,6 @@ var (
 	toLogLevel = toLogLevelFromString
 )
 
-// ------------------------------------------------------------------------------------------------
-// Custom GETs
-// ------------------------------------------------------------------------------------------------
```
torives, line 53: The documentation at the start of this file should be updated now that custom get functions are defined in the `custom_get.go` file.

internal/config/generate/main.go
```diff
@@ -14,7 +14,11 @@ import (
 
 // This script will read the Config.toml file and create:
 //
-// - a formatted get.go file, with get functions for each environment variable;
+// - a formatted get.go file, with  get functions for some environment variables;
+//
+// - a formated nodeconfig.go file with the Node Config Struct and two init functions
+// , one that initializes it reading the values from environent variables, and other that
+// initializes it with default values;
```
torives, line 21:
```suggestion
// - a formatted get.go file, with  get functions for some environment variables;
//
// - a formatted nodeconfig.go file with the NodeConfig struct and two init functions:
// one that initializes it by reading the values from environment variables and another that
// initializes it with default values;
```

internal/config/log.go
```diff
@@ -18,12 +18,13 @@ var (
 )
 
 func init() {
-	logInit()
+	nodeConfig := NewNodeConfigFromEnv()
+	logInit(nodeConfig)
 }
 
-func logInit() {
+func logInit(nodeConfig NodeConfig) {
```
torives, line 25:
```suggestion
func InitLog(nodeConfig NodeConfig) {
```
It's better to export this function and call it directly after validating the config in `cmd/cartesi-rollups-node/main.go`. Then, the `init()` function from line 20 can be removed.

internal/config/validate.go
```diff
@@ -0,0 +1,72 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+package config
+
+import "fmt"
+
+func (nodeConfig *NodeConfig) Validate() {
```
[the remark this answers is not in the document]
fmoura: I see that as an improvement, like checking the URL format, if "block number" is an actual number and things alike. These were not being check before. So I propose it being a new PR/issue, since this is already a "long PR"

internal/config/log.go
fmoura: Without the default config package `init()` , that initializes the logs, some tests may fail.  For those, the logs must be initialized at those tests

torives (changes requested):
After some consideration, I changed my mind about merging this PR and dealing with its implications in a follow-up issue (#325). 

As we've noticed multiple times, the old design of the `config` package doesn't fit our current needs, and adapting it will inevitably create technical debt, something we should avoid.

Even though we've already spent a great amount of time and effort with this, I'm convinced that going back to the drawing board and creating a design from scratch that better accounts for the new requirements is a better approach.

PR #327, opened 2024-02-22
feat: refactor node config

fmoura:


internal/config/generate/helpers.go
endersonmaia, line 1: Can't we use templates here? I think it would be easier to grasp how the final go code would look like, instead of building it with helper functions like `addLine` and such.

https://pkg.go.dev/text/template

internal/config/validate.go
```diff
@@ -0,0 +1,65 @@
...
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+package config
+
+import "fmt"
+
+func (nodeConfig *NodeConfig) Validate() {
+	if nodeConfig.cartesiExperimentalSunodoValidatorEnabled != nil &&
+		*nodeConfig.cartesiExperimentalSunodoValidatorEnabled {
+		if nodeConfig.cartesiExperimentalSunodoValidatorRedisEndpoint == nil {
+			panic("missing CartesiExperimentalSunodoValidatorRedisEndpoint config")
```
endersonmaia, line 11: I think it's better to inform with the format the configuration is defined:

```suggestion
			panic("CARTESI_EXPERIMENTAL_SUNODO_VALIDATOR_REDIS_ENDPOINT environment variable not defined")
```

endersonmaia (changes requested):

internal/config/generate/helpers.go
fmoura: I will leave this to a new PR. So we can move forward with other PRs that are on top of [PR #318](https://github.com/cartesi/rollups-node/pull/318)

PR #335, opened 2024-03-04
feat: Add bindings to emulator C API

mpernambuco:
Initial draft of the go bindings to the machine emulator C API


pkg/emulator/emulator/cartesi-machine-emulator.go
```diff
@@ -0,0 +1,967 @@
+package emulator
```
renan061, line 1: Go convention: this file can be renamed to just `emulator.go`.

pkg/emulator/emulator/cartesi-machine-emulator.go
```diff
@@ -0,0 +1,967 @@
...
+#include <stdlib.h>
+*/
+import "C"
+
+import (
+	"encoding/hex"
+	"unsafe"
+)
+
+type Error int32
```
renan061, line 16: Go convention: its best to return the `error` type.

Functions currently are: `func foo() (something *Something, errMsg string, errCode Error)`.
It would be best if they were: `func foo() (something *Something, err error)`.

Then, we could have a special struct type:

```
type Error struct {
	Code int32
	Msg string
}
```

That implements the `error` interface:

```
func (err Error) String() string {
	return "cartesi-machine error: " + err.Code.String() + " (" + err.Msg + ")"
}
```

The function that instantiates the `Error` struct and returns an `error` could be something like this:

```
func newError(code C.int, msg *C.char) error {
	defer C.cm_delete_cstring(msg)
	if code := mapCCodeToGoCode(C.CM_ERROR(code)); code != ErrorOk {
		return errors.New(Error{Code: code, Msg: C.GoString(message)}.String())
	}
	return nil
}
```

pkg/emulator/emulator/cartesi-machine-emulator.go
```diff
@@ -0,0 +1,967 @@
...
+
+const (
+	ERROR_OK                      Error = C.CM_ERROR_OK
```
renan061, line 19: These can then have type `ErrorCode`, given something like `type ErrorCode = uint32`.

pkg/emulator/emulator/cartesi-machine-emulator.go
```diff
@@ -0,0 +1,967 @@
...
+	cref          *C.cm_jsonrpc_mg_mgr
+}
+
+// a local or remote machine
+type Machine struct {
+	cref          *C.cm_machine
+	remoteManager *RemoteMachineManager
+}
+
+func (x *RemoteMachineManager) Free() {
```
renan061, line 582: Go convention: name the self instance something like `manager` instead of `x`.

pkg/emulator/emulator/cartesi-machine-emulator.go
```diff
@@ -0,0 +1,967 @@
...
+	if x != nil && x.cref != nil {
+		C.cm_delete_jsonrpc_mg_mgr(x.cref)
+		x.cref = nil
+	}
+}
+
+func (x *Machine) Free() {
```
renan061, line 589: Go convention: name the `Machine` self instance `machine` instead of `x`.

pkg/emulator/emulator/cartesi-machine-emulator.go
```diff
@@ -0,0 +1,967 @@
...
+		errCode = Error(e)
+		errMsg = C.GoString(cerr)
+		C.cm_delete_cstring(cerr)
+		return
+	}
+	return machine, "", ERROR_OK
+}
+
+func GetDefaultConfig() (config *MachineConfig, errMsg string, errCode Error) {
+	theriCfg := theirMachineConfigCRef{}
```
renan061, line 643: Typo: `theirCfg`.

pkg/emulator/emulator/cartesi-machine-emulator.go
```diff
@@ -0,0 +1,967 @@
...
+		errMsg = C.GoString(cerr)
+		C.cm_delete_cstring(cerr)
+		return errMsg, errCode
+	}
+	return "", ERROR_OK
+}
+
+type MerkleTreeHash [32]byte
+
+func (x *MerkleTreeHash) ToString() string {
```
renan061, line 842: Go convention: name this just `String`.

pkg/emulator/go.mod
```diff
@@ -0,0 +1,3 @@
+module go-emulator
```
renan061, line 1: This folder does not need a `.mod` file, because it is inside the larger Go module from the `.mod` file at the root.

pkg/emulator/main.go
```diff
@@ -0,0 +1,187 @@
+package main
```
renan061, line 1: Since we can't have a `.mod` file, we also can't have a main here.
So we need this to be a test, as in `emulator_test.go` instead of `main.go`.
And package must be `emulator`.

pkg/emulator/main.go
```diff
@@ -0,0 +1,187 @@
...
+import (
+	"flag"
+	"fmt"
+	"go-emulator/emulator"
+	"math"
+	"os"
+	"strings"
+)
+
+func main() {
```
renan061, line 12: Then this should be named `TestEmulator(t *testing.T)`.

You can use the `require` package to make assertions within the test.

pkg/emulator/main.go
```diff
@@ -0,0 +1,187 @@
...
+	var machine *emulator.Machine
+	defer machine.Free()
+	var mgr *emulator.RemoteMachineManager
+	defer mgr.Free()
+	var err emulator.Error
+	var errMsg string
+	runtimeConfig := &emulator.MachineRuntimeConfig{}
+
+	// Parse command line arguments
```
renan061, line 21: I don't know if these will work within a test. We could have global `const` variables instead.

renan061 (changes requested):
Great work on the binding, Perna! :-)

I left a lot of comments, but they are mostly style changes so the code conforms to Go conventions.

Also, we should write this binding using the output unification branch from the start

gligneul (approved):
I'm approving it, but first, we should fix the CI by installing the machine emulator.

PR #337, opened 2024-03-06
Add new GraphQL API schemas

GMKrieger:
This PR is meant to be a place to discuss the possible changes for the API. It will not be merged as is, but rather is a way to share this change.

The new schemas were generated with PostGraphile, based on the node database. For more information on PostGraphile, you can access https://www.graphile.org/postgraphile/.

All inputs are welcome!


gligneul:
Let's close this PR for the time being and focus on the database schema.

PR #339, opened 2024-03-08
Validate Chain ID

### LINKED ISSUE #245: Check RPC chain ID
## 📚 Context

The rollups node can fetch the chain ID from the Ethereum node RPC using the following method: https://ethereum-org-fork.netlify.app/developers/docs/apis/json-rpc#eth_chainid

Then, the node should compare the RPC chain ID with the configuration chain ID to make sure the node is executing in the correct chain.

torives:
Closes #245

### DIFF
--- a/cmd/cartesi-rollups-node/chainid.go
+++ b/cmd/cartesi-rollups-node/chainid.go
@@ -1,0 +1,46 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/cartesi/rollups-node/internal/config"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+const defaultTimeout = 3 * time.Second
+
+// Checks if the chain id from the configuration matches the chain id reported
+// by the Ethereum node. If they don't, it returns an error.
+func validateChainId(ctx context.Context, chainId uint64, ethereumNodeAddr string) error {
+	remoteChainId, err := getChainId(ctx, ethereumNodeAddr)
+	if err != nil {
+		config.ErrorLogger.Printf("Couldn't validate chainId: %v\n", err)
+	} else if chainId != remoteChainId {
+		return fmt.Errorf(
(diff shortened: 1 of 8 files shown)


cmd/cartesi-rollups-node/chainid.go
```diff
@@ -0,0 +1,73 @@
...
+			"chainId mismatch. Expected %v but Ethereum node returned %v",
+			chainId,
+			remoteChainId,
+		)
+	}
+	return nil
+}
+
+func getChainId(ethereumNodeAddr string) (uint, error) {
+	// create and send the request
```
gligneul, line 34: Why not use go-ethereum to do the call?

gligneul (changes requested):

cmd/cartesi-rollups-node/chainid.go
torives: There's no reason at all, I just forgot that was an option 🤦🏽‍♂️

cmd/cartesi-rollups-node/chainid.go
```diff
@@ -0,0 +1,46 @@
...
+
+	client, err := ethclient.Dial(ethereumNodeAddr)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to create RPC client: %v", err)
+	}
+	chainId, err := client.ChainID(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to get chain id: %v", err)
+	}
+	return uint(chainId.Uint64()), nil
```
gligneul, line 45: I don't feel comfortable with this cast. I think we should use uint64.

cmd/cartesi-rollups-node/chainid_test.go
```diff
@@ -0,0 +1,48 @@
...
+type ValidateChainIdSuite struct {
+	suite.Suite
+}
+
+func TestValidateChainId(t *testing.T) {
+	suite.Run(t, new(ValidateChainIdSuite))
+}
+
+func (s *ValidateChainIdSuite) TestItFailsIfChainIdsDoNotMatch() {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
gligneul, line 25: nice

gligneul (approved):

PR #350, opened 2024-03-13
Support specification of private key to claimer

### LINKED ISSUE #344: Support CARTESI_AUTH_PRIVATE_KEY
## 📚 Context

Sometimes you have a private key (hash of 32 bytes), but don't have a mnemonic.

## ✔️ Solution

Support specification of CARTESI_AUTH_PRIVATE_KEY, which is mutually exclusive with other AUTH configs, like CARTESI_AUTH_MNEMONIC, CARTESI_AUTH_MNEMONIC_FILE, CARTESI_AUTH_MNEMONIC_ACCOUNT_INDEX and CARTESI_AUTH_AWS_KMS_KEY_ID

## 📈 Subtasks

- [x] add config option

tuler:
This PR adds the option to specify CARTESI_AUTH_PRIVATE_KEY to pass to the authority-claimer in case the user have a simple private key in hand, and not a mnemonic.

The value of the private key must be a hex string of 32 bytes, without the traditional starting `0x`.
i.e.: `59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d`

closes #344

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -53,6 +53,14 @@ Overrides `CARTESI_AUTH_AWS_KMS_*`.
 
 * **Type:** `string`
 
+## `CARTESI_AUTH_PRIVATE_KEY`
+
+The node will use this private key to sign transactions.
+
+Overrides `CARTESI_AUTH_MNEMONIC`, `CARTESI_AUTH_MNEMONIC_FILE` and `CARTESI_AUTH_AWS_KMS_*`.
+
+* **Type:** `string`
+
 ## `CARTESI_BLOCKCHAIN_BLOCK_TIMEOUT`
 
 Block subscription timeout in seconds.
--- a/docs/config.md
+++ b/docs/config.md
@@ -57,6 +57,14 @@ Overrides `CARTESI_AUTH_AWS_KMS_*`.
 
 The node will use this private key to sign transactions.
 
+Overrides `CARTESI_AUTH_PRIVATE_KEY_FILE`, `CARTESI_AUTH_MNEMONIC`, `CARTESI_AUTH_MNEMONIC_FILE` and `CARTESI_AUTH_AWS_KMS_*`.
+
+* **Type:** `string`
+
+## `CARTESI_AUTH_PRIVATE_KEY_FILE`
+
+The node will use the private key contained in this file to sign transactions.
+
 Overrides `CARTESI_AUTH_MNEMONIC`, `CARTESI_AUTH_MNEMONIC_FILE` and `CARTESI_AUTH_AWS_KMS_*`.
 
 * **Type:** `string`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added verification to ensure CARTESI_BLOCKCHAIN_ID matches the id returned from the Ethereum node
-- Added support for CARTESI_AUTH_PRIVATE_KEY
+- Added support for CARTESI_AUTH_PRIVATE_KEY and CARTESI_AUTH_PRIVATE_KEY_FILE
 
 ## Changed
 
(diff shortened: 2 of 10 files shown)


endersonmaia (changes requested):
It should also be possible to alternatively define a `CARTESI_AUTH_PRIVATE_KEY_FILE`, as we have for `CARTESI_AUTH_MNEMONIC_FILE`.

See: https://github.com/cartesi/rollups-node/blob/main/internal/config/generate/Config.toml#L154-L162

tuler:
> It should also be possible to alternatively define a `CARTESI_AUTH_PRIVATE_KEY_FILE`, as we have for `CARTESI_AUTH_MNEMONIC_FILE`.
> 
> See: https://github.com/cartesi/rollups-node/blob/main/internal/config/generate/Config.toml#L154-L162

Done

docs/config.md
```diff
@@ -53,6 +53,22 @@ Overrides `CARTESI_AUTH_AWS_KMS_*`.
...
+
+Overrides `CARTESI_AUTH_PRIVATE_KEY_FILE`, `CARTESI_AUTH_MNEMONIC`, `CARTESI_AUTH_MNEMONIC_FILE` and `CARTESI_AUTH_AWS_KMS_*`.
+
+* **Type:** `string`
+
+## `CARTESI_AUTH_PRIVATE_KEY_FILE`
+
+The node will use the private key contained in this file to sign transactions.
+
+Overrides `CARTESI_AUTH_MNEMONIC`, `CARTESI_AUTH_MNEMONIC_FILE` and `CARTESI_AUTH_AWS_KMS_*`.
```
renan061, line 68: `Overrides `CARTESI_AUTH_MNEMONIC_*` and `CARTESI_AUTH_AWS_KMS_*`.`

docs/config.md
```diff
@@ -53,6 +53,22 @@ Overrides `CARTESI_AUTH_AWS_KMS_*`.
 
 * **Type:** `string`
 
+## `CARTESI_AUTH_PRIVATE_KEY`
+
+The node will use this private key to sign transactions.
+
+Overrides `CARTESI_AUTH_PRIVATE_KEY_FILE`, `CARTESI_AUTH_MNEMONIC`, `CARTESI_AUTH_MNEMONIC_FILE` and `CARTESI_AUTH_AWS_KMS_*`.
```
renan061, line 60: ```Overrides `CARTESI_AUTH_PRIVATE_KEY_FILE`, `CARTESI_AUTH_MNEMONIC_*`, and `CARTESI_AUTH_AWS_KMS_*`.```

docs/config.md
tuler: I think it's more clear the way I did instead of using `*`

docs/config.md
tuler: I think it's more clear the way I did instead of using `*`

docs/config.md
```diff
@@ -53,6 +53,22 @@ Overrides `CARTESI_AUTH_AWS_KMS_*`.
 
 * **Type:** `string`
 
+## `CARTESI_AUTH_PRIVATE_KEY`
+
+The node will use this private key to sign transactions.
```
endersonmaia, line 58: I suggest adding the format expected:

Ex.: CARTESI_AUTH_PRIVATE_KEY_FILE=59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d

endersonmaia (changes requested):

docs/config.md
tuler: No other option has example. So I didn't know how to add that.

docs/config.md
renan061: I was going to suggest to than remove the `*` from the `KMS_*`, but @gligneul told me this description is going to change in his `config` PR, so best to leave it as it is.

renan061 (approved):

endersonmaia (approved):

PR #356, opened 2024-03-15
Create runtime dir for container image

endersonmaia:
This PR will create the RUNTIME_DIR that wasn't being created before the permissions were being given with `chown`.

This will allow sunodo to deprecate the usage/maintenance of its own rollups-node container image.

Related to: https://github.com/sunodo/sunodo/pull/413


endersonmaia:
I propose we do this in a 1.3.2 release. 

Let me know if it's possible.

endersonmaia:
Not necessary.

`WORKDIR` already creates the directory

PR #357, opened 2024-03-19
Validate machine hash

### LINKED ISSUE #259: Check machine hash in Go

torives:
Closes #259.


torives:
The code can be reviewed, but this PR will be rebased on top of the main branch after #354 is merged.

torives:
This is now ready for review

build/compose-host.yaml
```diff
@@ -9,3 +9,4 @@ services:
       - "10007:10007" # Host Runner Rollup API
     environment:
       CARTESI_FEATURE_HOST_MODE: "true"
+      CARTESI_FEATURE_DISABLE_MACHINE_HASH_CHECK: "true"
```
gligneul, line 12: Please also add this to the `build/compose-devnet.yam`.

internal/node/machinehash_test.go
```diff
@@ -0,0 +1,156 @@
...
+func (s *ValidateMachineHashSuite) TestItFailsWhenSnapshotHasNoHash() {
+	machineDir, err := os.MkdirTemp("", "")
+	if err != nil {
+		s.FailNow(err.Error())
+	}
+
+	err = validateMachineHash(context.Background(), machineDir, "", "")
+	s.ErrorContains(err, "no such file or directory")
+
+	os.RemoveAll(machineDir)
```
gligneul, line 38: This should be a defer because if the test fails, this won't be called.

internal/node/machinehash_test.go
```diff
@@ -0,0 +1,156 @@
...
+func (s *ValidateMachineHashSuite) TestItFailsWhenHashHasWrongSize() {
+	machineDir, err := mockMachineDir("deadbeef")
+	if err != nil {
+		s.FailNow(err.Error())
+	}
+
+	err = validateMachineHash(context.Background(), machineDir, "", "")
+	s.ErrorContains(err, "wrong size")
+
+	os.RemoveAll(machineDir)
```
gligneul, line 50: Here as well.

internal/node/machinehash_test.go
```diff
@@ -0,0 +1,156 @@
...
+	err = validateMachineHash(
+		ctx,
+		machineDir,
+		addresses.GetTestBook().CartesiDApp.String(),
+		BlockchainHttpEndpoint,
+	)
+	s.NotNil(err)
+	s.ErrorIs(err, context.DeadlineExceeded)
+
+	os.RemoveAll(machineDir)
```
gligneul, line 74: Here as well.

internal/node/machinehash_test.go
```diff
@@ -0,0 +1,156 @@
...
+	err = deps.Terminate(context.Background(), devnet)
```
gligneul, line 75: Here as well?

internal/node/machinehash_test.go
```diff
@@ -0,0 +1,156 @@
...
+		ctx,
+		machineDir,
+		addresses.GetTestBook().CartesiDApp.String(),
+		BlockchainHttpEndpoint,
+	); err != nil {
+		s.FailNow(err.Error())
+	}
+
+	os.RemoveAll(machineDir)
+	err = deps.Terminate(ctx, devnet)
```
gligneul, line 102: Here as well.

gligneul (changes requested):

gligneul (approved):

internal/node/machinehash_test.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		s.FailNow(err.Error())
+	}
+	devnet, err := startDevnet()
+	if err != nil {
+		s.Fail(err.Error())
+		if err = deps.Terminate(context.Background(), devnet); err != nil {
+			s.Fail(err.Error())
+		}
+	}
```
gligneul, line 65: can you try something like that?
```suggestion
	s.Require().Nil(err)
	defer func() {
		err := deps.Terminate(context.Background(), devnet)
		s.Require().Nil(err)
	}()
```

torives: I just realized it made no sense in the first place since the failing function was the one starting the containers, so there's no container to stop after all.

gligneul: The defer recommendation still stands.

gligneul (approved):

PR #358, opened 2024-03-22
feat: Add Echo App end-to-end test

fmoura:
to run the end to end tests you should run : 

```
GOFLAGS="-count=1" go test ./test/ --tags=endtoendtests
```

If you want to have a glimpse of the logs use `-v` arg :
```
GOFLAGS="-count=1" go test ./test/ --tags=endtoendtests -v 
```

And as it uses the "natively run" mode, every dependency must be correctly installed before running these


end-to-end-tests/client_util.go
```diff
@@ -0,0 +1,60 @@
...
+	payloadBytes, _ := hexutil.Decode(payload)
+	inputIndex, err := ethutil.AddInput(ctx, client, book, signer, payloadBytes)
+	if err != nil {
+		return 0, err
+	}
+	return inputIndex, nil
+
+}
+
+func AdvanceTime(ctx context.Context, blockchainHttpEnpoint string, timeInSeconds int) error {
```
fmoura, line 41: Just found out that `cartesi-rollups-cli` has another implementation of this, using JSON RCP instead of Go Etherium. We should have only one

fmoura: Should I change the Command to use this one? and move this util functions to `ethutil.go` ?

end-to-end-tests/echo_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+}
+
+func (s *EchoInputTestSuite) SetupTest() {
+
+}
+
+func (s *EchoInputTestSuite) TestSendInput() {
+
+	ctx := context.Background()
```
gligneul, line 128: We should create a WithTimeout context to avoid the chance of having the test hanging.

end-to-end-tests/echo_test.go
```diff
@@ -0,0 +1,205 @@
...
+	require.EqualValues(s.T(), 0, inputIndex)
+
+	require.NoError(s.T(), AdvanceTime(ctx, LocalBlockchainHttpEndpoint, 120))
+
+	// Get Input with vouchers and proofs
+	graphQlClient := graphql.NewClient("http://localhost:10000/graphql", nil)
+	getInputChan := make(chan *readerclient.Input, 1)
+	getInputErr := make(chan struct{}, 1)
+
+	go func() {
```
gligneul, line 143: I didn't understand why this needed to be a goroutine.

end-to-end-tests/echo_test.go
```diff
@@ -0,0 +1,205 @@
...
+		require.NoError(s.T(), err)
+		break
+	case <-getInputErr:
+		require.FailNow(s.T(), "Could not retrieve Voucher with Proof 0")
+	}
+
+}
+
+func readExpectedInputWithProof(filename string) (*readerclient.Input, error) {
+	cwd, err := os.Getwd()
```
gligneul, line 182: You can use the embed package to add these files to the node binary. See https://pkg.go.dev/embed

gligneul (changes requested):

end-to-end-tests/echo_test.go
fmoura: The Go Test itself has a timeout feature. The default is 10 minutes.

end-to-end-tests/echo_test.go
```diff
@@ -0,0 +1,205 @@
...
+	machine.Save("cartesi/rollups-node-snapshot:devel", tempDir, "test-echo-app")
+
+	// Run deps
+	ctx := context.Background()
+	var depsConfig = deps.NewDefaultDepsConfig()
+	depsContainers, err := deps.Run(ctx, *depsConfig)
+	require.NoError(s.T(), err)
+
+	// Wait a little for Dependencies to be actually ready
+	time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
fmoura: Despite the Devnet container being healthy, not waiting a little more, yield an error as the block "first" block is not available. Actually this should be reolved/corrected inside deps.go and not here.

end-to-end-tests/echo_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// Fix the Blochain timestamp. Must be "in the future"
+	err = SetNextBlockTimestamp(ctx, LocalBlockchainHttpEndpoint, 7000000000)
```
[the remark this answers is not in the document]
fmoura: e2e_tests, maybe ?

end-to-end-tests/echo_test.go
fmoura: The idea was that the following `select` would wait for what happens first : input being correctly retrieved or supervisor giving an error. 

Even if the node died, the loop will continue until it reaches the `maxReadInputAttemps`. 

But as it does not takes that long to execute, the necessity of being a goroutine is very low.

end-to-end-tests/echo_test.go
```diff
@@ -0,0 +1,205 @@
...
+	supervisorErr chan error
+}
+
+func (s *EchoInputTestSuite) SetupSuite() {
+
+	// Clear default_tx_database
+	cwd, err := os.Getwd()
+	require.NoError(s.T(), err)
+
+	_ = os.Remove(filepath.Join(cwd, defaulTxDatabaseFile))
```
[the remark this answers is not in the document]
fmoura: Will try again to run tests on a temp dir. Snapshot is already being saved on a temp dir and removed when tests ends

end-to-end-tests/echo_test.go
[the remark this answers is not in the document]
fmoura: renamed it to test after all

end-to-end-tests/echo_test.go
fmoura: Made the context to have timeout also, despite the tests already having a timeout feature

CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added verification to ensure the Cartesi Machine snapshot hash matches the template hash from the CartesiDApp contract
 - Added support for `CARTESI_AUTH_PRIVATE_KEY` and `CARTESI_AUTH_PRIVATE_KEY_FILE`
 - Added `CARTESI_AUTH_KIND` environment variable to select the blockchain authetication method
-- Added structured logging with slog. Colored logs can now be enabled with `CARTESI_LOG_PRETTY` environment variable.
+- Added structured logging with slog. Colored logs can now be enabled with `CARTESI_LOG_PRETTY` environment variable
+- Added Rollps end-to-end test using Echo Dapp
```
gligneul, line 18: typo

end-to-end-tests/config.go
```diff
@@ -0,0 +1,85 @@
...
+	LocalInputBoxDeploymentBlockNumber    = 20
+	LocalHttpAddress                      = "0.0.0.0"
+	LocalHttpPort                         = 10000
+	LocalBlockTimeout                     = 60
+	LocalFinalityOffset                   = 1
+	LocalEpochDurationInSeconds           = 120
+)
+
+func NewLocalNodeConfig() config.NodeConfig {
+	var nodeConfig config.NodeConfig
```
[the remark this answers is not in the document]
fmoura: This method is intended to be used only for the endtoendtests package, maybe the default should be Debug. Lint is complaining this method is not used, despite it is used in the tests, when it is not public

gligneul (approved):

gligneul (approved):

internal/services/service.go
```diff
@@ -14,5 +14,5 @@ type Service interface {
 	fmt.Stringer
 
 	// Starts a service and sends a message to the channel when ready
-	Start(ctx context.Context, ready chan<- struct{}) error
+	Start(ctx context.Context, ready chan<- struct{}, workDir string) error
```
torives, line 17: I don't like this new parameter on `service.Start()` because the concept doesn't apply to every service, only to `CommandService`. I suggest adding `workDir` as a property of `CommandService` and propagating it downwards from `node.Setup()` to the relevant functions.

The main rationale is that these methods will either die or change in node v2.0 since services will become Go code, no longer Rust executables, but the Service interface will remain and I would like to keep it correct.

torives (changes requested):

internal/services/service.go
fmoura: Done!

torives (approved):

gligneul (approved):

Issue #347, opened 2024-03-13
Create base database migrations


GMKrieger:
## 📚 Context

The new database designed for the output unification will be remade in go and needs setup correctly during startup.

## ✔️ Solution

Add database migrations to the node that guarantee that the schema is consistent when we run the node. This first step will contain the basic tables agreed to be necessary for the continued development of the other services.

## 📈 Subtasks

- [ ] Study and choose a suitable library for the database in Go
- [ ] Implement base migrations for the database
- [ ] Implement fundamental queries for the database
- [ ] Create tests for new repository

PR #365, opened 2024-03-26
Add database migrations

GMKrieger:
#347


internal/data/migrations.go
```diff
@@ -0,0 +1,33 @@
...
+	_ "github.com/golang-migrate/migrate/v4/database/postgres"
+	_ "github.com/golang-migrate/migrate/v4/source/file"
+	"github.com/golang-migrate/migrate/v4/source/iofs"
+)
+
+//go:embed migrations/*
+var content embed.FS
+
+func RunMigrations(postgres_endpoint string) {
+	d, err := iofs.New(content, "migrations")
```
torives, line 21: I'd rather use a more descriptive name for the `d` variable. IMO the only context where they're okay is when defining member functions.

internal/data/migrations.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		slog.Error("Unable to use embed files: ", err)
+	}
+
+	m, err := migrate.NewWithSourceInstance("iofs", d, postgres_endpoint)
```
torives, line 26: The same goes for `m` here, let's give it a proper name.

torives (commented):
It is not a dealbreaker but I want to register my vote to rename the package to `repository`. It's more descriptive than `data` and it's already the name of the main file, so... 😉

internal/data/migrations/000001_create_inputs_outputs_reports_proofs.up.sql
```diff
@@ -0,0 +1,48 @@
...
+    CONSTRAINT "reports_input_index_fkey" FOREIGN KEY ("input_index") REFERENCES "inputs"("index")
+);
+
+CREATE TABLE "proofs"
+(
+    "input_index" INT NOT NULL,
+    "output_index" INT NOT NULL,
+    "first_input_index" INT NOT NULL,
+    "last_input_index" INT NOT NULL,
+    "validity_input_index_within_epoch" INT NOT NULL,
```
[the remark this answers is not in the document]
GMKrieger: The input index within epoch is the value of the index counted from the first input from the epoch

`inputIndex = inputRange.firstIndex + inputIndexWithinEpoch`

internal/repository/repository_test.go
```diff
@@ -0,0 +1,131 @@
...
+	input := Input{
+		Index:  0,
+		Status: "UNPROCESSED",
+		Blob:   common.Hex2Bytes("deadbeef"),
+	}
+
+	err := s.database.InsertInput(s.ctx, &input)
+	if !s.Nil(err) {
+		s.T().FailNow()
+	}
```
torives, line 64:
```suggestion
	s.Require().Nil(err)
```

internal/repository/repository_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	response, err := s.database.GetInput(s.ctx, 0)
+	if !s.Nil(err) {
+		s.T().FailNow()
+	}
```
torives, line 69:
```suggestion
	s.Require().Nil(err)
```

internal/repository/repository_test.go
```diff
@@ -0,0 +1,131 @@
...
+	output := Output{
+		Index:      0,
+		InputIndex: 1,
+		Blob:       common.Hex2Bytes("deadbeef"),
+	}
+
+	err = s.database.InsertOutput(s.ctx, true, &output)
+	if !s.Nil(err) {
+		s.T().FailNow()
+	}
```
torives, line 96:
```suggestion
	s.Require().Nil(err)
```

internal/repository/repository_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	response, err := s.database.GetOutput(s.ctx, true, 1, 0)
+	if !s.Nil(err) {
+		s.T().FailNow()
+	}
```
torives, line 101:
```suggestion
	s.Require().Nil(err)
```

internal/repository/repository_test.go
```diff
@@ -0,0 +1,131 @@
...
+		postgres.WithUsername(dbUser),
+		postgres.WithPassword(dbPassword),
+		testcontainers.WithWaitStrategy(
+			wait.ForLog("database system is ready to accept connections").
+				WithOccurrence(2).
+				WithStartupTimeout(5*time.Second)),
+	)
+
+	return container, err
+
```
torives, line 130:
```suggestion

```

internal/repository/repository_test.go
```diff
@@ -0,0 +1,131 @@
...
+type RepositorySuite struct {
+	suite.Suite
+	ctx      context.Context
+	cancel   context.CancelFunc
+	postgres *postgres.PostgresContainer
+	database *database
+}
+
+func (s *RepositorySuite) SetupSuite() {
+	s.ctx, s.cancel = context.WithTimeout(context.Background(), testTimeout)
```
torives, line 30: I get the safety preoccupation, but I'm not a fan of the hard-coded timeout. IMO such a timeout should be configured at the `go test` level (especially on the CI), as some machines will naturally take longer than others.

torives: Loved the setup/teardown, though 👌🏽

internal/repository/repository_test.go
```diff
@@ -0,0 +1,131 @@
...
+	input := Input{
+		Index:  1,
+		Status: "UNPROCESSED",
+		Blob:   common.Hex2Bytes("deadbeef"),
+	}
+
+	err := s.database.InsertInput(s.ctx, &input)
+	if !s.Nil(err) {
+		s.T().FailNow()
+	}
```
torives, line 85:
```suggestion
	s.Require().Nil(err)
```

torives (commented):
I'd love to see more tests going over the failure scenarios: trying to get the second output of an input that only has one, that sort of thing.

internal/repository/repository_test.go
GMKrieger: I've just used the same configuration as used on the eth_util tests. I think, for consistency, we keep them the same.

Aside from that, there's a a discussion to be had that a 5 minute timeout is good actually, because if the test is taking that long, we really should raise an alarm as to something not being right.

internal/repository/repository_test.go
```diff
@@ -0,0 +1,159 @@
...
+	s.Require().Nil(err)
+
+	s.Require().Equal(&input, response)
+}
+
+func (s *RepositorySuite) TestInputFailsDoesntExist() {
+	response, err := s.database.GetInput(s.ctx, 1)
+	s.Require().Nil(response)
+
+	s.Require().Contains(fmt.Sprintf("%v", err), "no rows in result set")
```
torives, line 72: It seems that `ErrorContains` is the function you wanted 😉 
```suggestion
	s.Require().ErrorContains(err, "no rows in result set")
```

torives (approved):

PR #385, opened 2024-04-16
feat: Add Cartesi Machine installation to CI

fmoura:


fmoura:
The Takeaways from this PR are: 

Given that :
- Cartesi Machine currently needs a base library only available on Debian Bookworm 
- Github Actiomn VMs only provides , as of today, Ubuntu 22.04 and Ubuntu 20.04 which are not Debian Bookworm based

The way to fully run node with a native Cartesi Machine , and in the future, to build Node with Machine SDK dependencies is to use a custom crafted Debian Bookworm Container, instead of Github Action VM directly. 

The only drawback is that , when using a container instead of VM, some of the Rust Code Tests fails. These should run directly on Github VM

Golang code tests run perfectly.

PR #386, opened 2024-04-18
Add Validator Mock

### LINKED ISSUE #369: Create Mocked Validator
## 📚 Context
Now that we have a general design for the database schema and how the Validator will interact with it, it is time to exercise the component's main logic.

## ✔️ Solution
To create a prototype to validate the Validator's main logic and APIs, mocking the database and the proof-generating code.

## 📈 Subtasks
- [x] Create the necessary types
- [x] Implement the main logic
- [x] Test main logic with mocks

torives:
Closes #369

### DIFF
--- a/internal/node/model/models.go
+++ b/internal/node/model/models.go
@@ -1,7 +1,7 @@
 // (c) Cartesi and individual authors (see AUTHORS)
 // SPDX-License-Identifier: Apache-2.0 (see LICENSE)
 
-package repository
+package model
 
 import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
--- a/internal/node/model/models.go
+++ b/internal/node/model/models.go
@@ -9,7 +9,7 @@ import (
 
 type Input struct {
 	// Input index starting from genesis
-	Index int `json:"index"`
+	Index uint64 `json:"index"`
 	// Status of the input
 	Status string `json:"status"`
 	// Input data as a blob, starting with '0x'
@@ -18,9 +18,9 @@ type Input struct {
 
 type Output struct {
 	// Input whose processing produced the output
-	InputIndex int `json:"inputIndex"`
+	InputIndex uint64 `json:"inputIndex"`
 	// Output index within the context of the input that produced it
-	Index int `json:"index"`
+	Index uint64 `json:"index"`
 	// Output data as a blob, starting with '0x'
 	Blob hexutil.Bytes `json:"blob"`
 }
--- a/internal/node/model/models.go
+++ b/internal/node/model/models.go
@@ -24,3 +24,29 @@ type Output struct {
 	// Output data as a blob, starting with '0x'
 	Blob hexutil.Bytes `json:"blob"`
 }
+
+type Epoch struct {
+	StartBlock uint64
+	EndBlock   uint64
+}
+
+type InputRange struct {
+	First uint64
+	Last  uint64
+}
+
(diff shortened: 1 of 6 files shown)


internal/node/model/models.go
```diff
@@ -0,0 +1,52 @@
...
+	// Output data as a blob, starting with '0x'
+	Blob hexutil.Bytes `json:"blob"`
+}
+
+type Epoch struct {
+	StartBlock uint64
+	EndBlock   uint64
+}
+
+type InputRange struct {
```
GMKrieger, line 33: I have some doubts about the need to have an InputRange struct. It isn't a "thing" by itself in the architecture, it's a value related to the claim. In your mock, the use of the struct aids in appending the values, but also adds a lot of lines to access these values.

I don't really have an alternative, just food for thought.

internal/validator/validator.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// generateProofs will create the proofs for all Outputs within an Epoch.
+// It returns the proofs and the root hash of the Merkle tree used to generate them
+func (v Validator) generateProofs(
+	ctx context.Context,
+	inputRange InputRange,
+	machineStateHash hexutil.Bytes,
+	outputs []Output,
+) ([]Proof, hexutil.Bytes, error) {
```
GMKrieger, line 196: Do we need to return the OutputsEpochRootHash as a different value? Can't we get it from any of the proofs?

internal/validator/validator_test.go
```diff
@@ -0,0 +1,292 @@
...
+		errChannel <- validator.Start(ctx, ready)
+	}()
+	cancel()
+
+	err := <-errChannel
+	s.Require().Equal(context.Canceled, err, "stopped for the wrong reason")
+}
+
+func (s *ValidatorSuite) TestItEventuallyBecomesReady() {
+	validator := NewValidator(s.repo, uint64(10), uint64(0))
```
GMKrieger, line 66: To be consistent with the other tests

```
	epochDuration := uint64(10)
	inputBoxDeploymentBlock := uint64(0)
	validator := NewValidator(s.repo, epochDuration, inputBoxDeploymentBlock)
```

GMKrieger (commented):
Just some minor comments

internal/validator/validator.go
```diff
@@ -0,0 +1,202 @@
...
+	repo ValidatorRepository,
+	epochDuration uint64,
+	inputBoxDeploymentBlock uint64,
+) Validator {
+	return Validator{repo, epochDuration, inputBoxDeploymentBlock}
+}
+
+// generateProofs will create the proofs for all Outputs within an Epoch.
+// It returns the proofs and the root hash of the Merkle tree used to generate them
+func (v Validator) generateProofs(
```
gligneul, line 191: I would make this a stand-alone function and put it in another file.

internal/node/model/models.go
torives: These are valid points and I agree with you: it isn't needed in any way. 

The `InputRange` struct was a late addition born from my intention to create familiarity between the node's code and the on-chain contracts (at least regarding the main entities) and thus ease understanding. IMO the tradeoff is worth it.

internal/validator/validator.go
torives: We can. I'll remove it from the returned values.

gligneul (approved):

PR #398, opened 2024-04-30
Feature/e2e test ci

fmoura:
Add End to End Tests to CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -126,8 +126,58 @@ jobs:
       - name: Run tests
         run: cargo test
 
+  build-ci-base:
+    runs-on: ubuntu-22.04
+    outputs:
+      output: ${{ steps.export_tag.outputs.image_tag }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+  
+      - name: Docker meta
+        id: docker_meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            name=ghcr.io/cartesi/rollups-node-ci-base
+          tags: |
+            type=semver,pattern={{version}}
+            type=ref,event=branch
+            type=ref,event=pr
+  
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+  
+      - uses: depot/setup-action@v1
+      - name: Build and push docker image
+        id: docker_build
+        uses: depot/bake-action@v1
+        with:
+          files: |
+            ./docker-bake.hcl
+            ${{ steps.docker_meta.outputs.bake-file }}
+            ./docker-bake.platforms.hcl
+          targets: rollups-node-ci-base
+          push: true
+          project: ${{ vars.DEPOT_PROJECT }}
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -126,8 +126,58 @@ jobs:
       - name: Run tests
         run: cargo test
 
+  build-ci-base :
```
gligneul, line 129:
```suggestion
  build-ci-base:
```

build/docker-bake.hcl
```diff
@@ -45,3 +45,10 @@ target "rollups-node-devnet" {
   inherits = ["common"]
   target   = "rollups-node-devnet"
 }
+
+target "rollups-node-ci-base" {
+  inherits = ["common"]
+  target   = "rollups-node-ci-base"
+  dockerfile = "./Dockerfile"
+  context  = "."
```
gligneul, line 53: Why overwrite these?

build/Dockerfile
```diff
@@ -331,3 +331,23 @@ USER cartesi
 
 # Set the Go supervisor as the command.
 CMD [ "cartesi-rollups-node" ]
+
+
+# STAGE: rollups-ci-base
+#
+# This stage prepares the base CI image
+FROM debian:bookworm as rollups-node-ci-base
```
gligneul, line 339:
```suggestion
FROM ${BASE_IMAGE} as rollups-node-ci-base
```

I would use the base image to be consistent.

build/Dockerfile
```diff
@@ -331,3 +331,23 @@ USER cartesi
...
+
+# Install git
+RUN apt update &&  apt install -y git-all
+
+# Install docker
+RUN <<EOF
+    apt update && apt install -y --no-install-recommends ca-certificates curl
```
gligneul, line 346: Why these two commands are separate?

.github/workflows/build.yml
```diff
@@ -152,6 +202,9 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: 'go.mod'
+      
+      - name: Fix VCS Linter Issue
```
gligneul, line 206: What is VCS? What is the issue?

.github/workflows/build.yml
```diff
@@ -126,8 +126,58 @@ jobs:
...
+      - name: Build and push docker image
+        id: docker_build
+        uses: depot/bake-action@v1
+        with:
+          files: |
+            ./docker-bake.hcl
+            ${{ steps.docker_meta.outputs.bake-file }}
+            ./docker-bake.platforms.hcl
+          targets: rollups-node-ci-base
+          push: true
```
gligneul, line 166: You should add this image to the clean-up job.
https://github.com/cartesi/rollups-node/blob/74d4381f84fb68ee9d6aa2baa9a3bff436b34b3f/.github/workflows/clean-up-images.yml#L19

gligneul (changes requested):

.github/workflows/build.yml
fmoura: Go Linter tries to check some "build stamping" against some information from the VCS, but fails reading this information.

build/Dockerfile
fmoura: Not really needed.

gligneul (approved):

PR #399, opened 2024-05-02
fix: fix Rust version at 1.77.0

fmoura:
Fix rust version at 1.77.0 to prevent unwanted issues. Rust version should be updated alongside Node Docker image generation


marcelstanley (commented):
Perhaps we should set this as a `chore` commit, @fmoura.

gligneul (approved):

PR #429, opened 2024-05-24
feat: normalize bool config parameters

### LINKED ISSUE #427: Normalize naming of boolean configuration parameters
## 📚 Context

Configuration parameters of type `bool` are not normalized (see [config.md for 2.0](https://github.com/cartesi/rollups-node/blob/next/2.0/docs/config.md))

## ✔️ Solution

As a suggestion by @endersonmaia captured at https://github.com/cartesi/rollups-node/issues/426#issuecomment-2127655208, we could use suffixes `ENABLED`/`DISABLED` for all `bool` arguments to imply state and normalize their meaning.

## 📈 Subtasks

- [x] Evaluate possible solutions along with the one listed above
- [x] Apply changes and update docs
- [x] Execute unit tests

marcelstanley:
Applied suffixes ENABLED/DISABLED to boolean parameters to match the node internal state.

Closes #427

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Bumped Rollups Contracts to 2.0
+- Normalized boolean configuration parameters (`CARTESI_LEGACY_BLOCKCHAIN_ENABLED`, `CARTESI_FEATURE_CLAIMER_ENABLED`, `CARTESI_FEATURE_MACHINE_HASH_CHECK_ENABLED`, `CARTESI_EXPERIMENTAL_SERVER_MANAGER_LOG_BYPASS_ENABLED` and `CARTESI_LOG_PRETTY_ENABLED`) and adjusted their logic accordingly
 
 ### Removed
 
--- a/cmd/cartesi-rollups-node/main.go
+++ b/cmd/cartesi-rollups-node/main.go
@@ -37,7 +37,7 @@ func main() {
 	opts := &tint.Options{
 		Level:      config.LogLevel,
 		AddSource:  config.LogLevel == slog.LevelDebug,
-		NoColor:    !config.LogPretty || !isatty.IsTerminal(os.Stdout.Fd()),
+		NoColor:    !config.LogPrettyEnabled || !isatty.IsTerminal(os.Stdout.Fd()),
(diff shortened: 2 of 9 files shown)


endersonmaia:
I thought `true` could be the default value for all those flags.

In a sense, if `CARTESI_LOG_PRETTY` being disabled is the default, then the variable should be named like `CARTESI_LOG_PRETTY_DISABLED=true`.

This way, if one needs to enable it, it will be necessary to explicitly set this to false, like `CARTESI_LOG_PRETTY_DISABLED=false`.

Also, if `CARTESI_FEATURE_MACHINE_HASH_CHECK` enabled is the default, then the variable should be `CARTESI_FEATURE_MACHINE_HASH_CHECK_ENABLE=true`, and if one needs to disable it, it will be required to explicitly set it to false like `CARTESI_FEATURE_MACHINE_HASH_CHECK_ENABLE=true`.

Does that make sense?

If yes, we'd have:

- `CARTESI_LEGACY_BLOCKCHAIN_DISABLED=true`
- `CARTESI_FEATURE_CLAIMER_DISABLED=true`
- `CARTESI_FEATURE_MACHINE_HASH_CHECK_ENABLED=tue`
- `CARTESI_EXPERIMENTAL_SERVER_MANAGER_LOG_BYPASS_DISABLED=true`
- `CARTESI_LOG_PRETTY_DISABLED=true`

torives (commented):
This can be quite contentious because you know, naming things is hard and mostly opinion-based. But my five cents is that we should avoid at all costs the cognitive load of dealing with double negatives.

`!LogPrettyDisabled` is harder to reason about than `!LogPrettyEnabled` as you have to consider the negation sign AND the variable name.

I propose to name them all with `_ENABLED` and set the default accordingly. We don't get the shared convention were all boolean vars are true or false by default, but we also don't have to deal with double negatives in the code (which is more important, IMO).

endersonmaia:
Yeah, I agree with you now that you make the point about the code.

I'm biased about configuring and running the node, and less about
maintaining the code itself.

Em ter., 11 de jun. de 2024, 20:47, Victor Yves ***@***.***>
escreveu:

torives:
> I'm biased about configuring and running the node, and less about maintaining the code itself.

@endersonmaia Which is perfectly natural. That is why your opinion is important, we get a different perspective and try to meet halfway 😉

torives (approved):

endersonmaia (approved):
LGTM :rocket:

PR #444, opened 2024-06-12
Add Merkle tree implementation in Go

### LINKED ISSUE #395: Implement the Merkle proof algorithm
## 📚 Context
The Merkle tree is the data structure that generates proofs for the outputs. It currently has three different implementations: 
- [Solidity](https://github.com/cartesi/rollups-contracts/blob/main/contracts/library/LibMerkle32.sol): used by the rollups contracts
- [C](https://github.com/cartesi/machine-emulator-tools/blob/main/sys-utils/libcmt/src/merkle.c): used inside the Cartesi Machine
- [Rust](https://github.com/cartesi/rollups-node/blob/main/offchain/host-runner/src/merkle_tree/complete.rs): used by the Node v1.x

The original plan for Node v2 was to port the Rust implementation to Go. This was later abandoned as the Rust implementation was deeply intertwined with the old way of generating proofs (dividing the original keccak256 hash into four 64-bit pieces and creating a Merkle tree with them), overcomplicating it.

## ✔️ Solution
The Node's use case for Merkle proofs is straightforward: it only needs to generate proofs at an epoch's end. By then, we already have all the outputs available and can calculate the root hash and the sibling arrays. This only needs to be done once, as there won't be any new outputs for that epoch that would require recalculating it.

Therefore, we can design a more efficient algorithm than a complete, generic Merkle tree data structure that wouldn't fit our current needs as well. The algorithm in question is very similar to the one currently implemented in Solidity.

## 📈 Subtasks
- [x] Write the algorithm in golang
- [x] Test it against the Solidity implementation
- [x] Test it against the C implementation

torives:
Closes #395

### DIFF
--- a/internal/merkle/proof.go
+++ b/internal/merkle/proof.go
@@ -1,0 +1,84 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package merkle
+
+import (
+	"fmt"
+
+	"github.com/cartesi/rollups-node/internal/node/model"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// CreateProofs creates proofs for all the leaves of a binary Merkle tree
+// with the given height. It returns the root hash and the siblings matrix
+// encoded as an array, in bottom-up order.
+//
+// If the number of leaves exceeds the capacity for the given height,
+// an error is returned.
+func CreateProofs(leaves []model.Hash, height uint) (model.Hash, []model.Hash, error) {
+	pristineNode := model.Hash{}
+
+	currentLevel := leaves
+	leafCount := uint(len(leaves))
+	siblings := make([]model.Hash, leafCount*height)
+
+	// for each level in the tree, starting from the leaves
+	for levelIdx := uint(0); levelIdx < height; levelIdx++ {
+		// for each leaf
+		for leafIdx := uint(0); leafIdx < leafCount; leafIdx++ {
+			// calculate its sibling at the current level
+
+			// Each pair of siblings shares the same parent. For any given parent with index i,
+			// the index of its left node is 2*i, and the index of its right node is 2*i+1.
+			// Since their indices differ only by the LSB, by flipping the LSB with ^1,
+			// we calculate the index of one another. In order to get the index of a parent,
+			// you remove the LSB with >>1. For an ancestor of height difference h,
+			// you remove the h LSBs with >>h.
+			siblingIdx := (leafIdx >> levelIdx) ^ 1
+			siblings[leafIdx*height+levelIdx] = *at(currentLevel, siblingIdx, &pristineNode)
+		}
+		// calculate the next level
+		currentLevel = parentLevel(currentLevel, &pristineNode)
+		// update the pristine node for the next level
+		pristineNode = crypto.Keccak256Hash(pristineNode[:], pristineNode[:])
+	}
+
+	// in the end, current level is the root level
+	if len(currentLevel) > 1 {
+		err := fmt.Errorf("too many leaves [%d] for height [%d]", leafCount, height)
+		return model.Hash{}, nil, err
+	}
+
+	return *at(currentLevel, 0, &pristineNode), siblings, nil
+}
+
+// parentLevel calculates the next level in a binary Merkle tree.
+//
+// For each pair of nodes in the current level, it computes their parent node
+// using the Keccak256 hashing algorithm. If level has an odd number of nodes,
+// a pristine node will be used to complete the pair.
+//
+// The parent nodes are stored in the first half of the original level slice.
+//
+// The function returns the parent level by re-slicing the original level slice.
+func parentLevel(level []model.Hash, pristineNode *model.Hash) []model.Hash {
+	// for each pair of nodes in level
+	for idx := 0; idx < len(level); idx += 2 {
+		leftChild := level[idx][:]
+		rightChild := at(level, uint(idx+1), pristineNode)[:]
+		// compute the parent node in-place
+		level[idx/2] = crypto.Keccak256Hash(leftChild, rightChild)
+	}
+	// return the parent level by re-slicing level
+	return level[:(len(level)+1)/2]
+}
+
+// at returns a pointer to the item located at index or the default value.
+func at(array []model.Hash, index uint, defaultValue *model.Hash) *model.Hash {
+	if index < uint(len(array)) {
+		return &array[index]
+	} else {
+		return defaultValue
+	}
+}
--- a/internal/merkle/proof_test.go
+++ b/internal/merkle/proof_test.go
@@ -1,0 +1,877 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package merkle
+
+import (
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/cartesi/rollups-node/internal/node/model"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/stretchr/testify/suite"
+)
+
+type CreateProofsSuite struct {
+	suite.Suite
+	pristine []model.Hash
+}
+
+func TestCreateProofsSuite(t *testing.T) {
+	suite.Run(t, new(CreateProofsSuite))
+}
+
+func (s *CreateProofsSuite) SetupSuite() {
+	maxHeight := 4
+	s.pristine = make([]model.Hash, maxHeight)
+
+	for height := 1; height < maxHeight; height++ {
+		s.pristine[height] = crypto.Keccak256Hash(
+			s.pristine[height-1][:],
+			s.pristine[height-1][:],
+		)
+	}
+}
+
+func (s *CreateProofsSuite) TestZeroHeight() {
+	s.Run("no leaves", func() {
+		root, siblings, err := CreateProofs(nil, 0)
+		s.Require().Nil(err)
+
+		s.Equal(s.pristine[0], root)
+		// if there are no leaves, there is nothing to prove
+		s.Equal(0, len(siblings))
+	})
+
+	s.Run("one leaf", func() {
+		leaf := crypto.Keccak256Hash([]byte("Cartesi"))
+
+		root, siblings, err := CreateProofs([]model.Hash{leaf}, 0)
+		s.Require().Nil(err)
+
+		s.Equal(leaf, root)
+		s.Equal(0, len(siblings))
+	})
+
+	s.Run("two leaves", func() {
+		leaves := make([]model.Hash, 2)
+
+		_, _, err := CreateProofs(leaves, 0)
+		s.Require().NotNil(err)
+
+		s.ErrorContains(err, "too many leaves")
+	})
+}
+func (s *CreateProofsSuite) TestHeightOne() {
+	height := 1
+	leaf1 := crypto.Keccak256Hash([]byte("Cartesi"))
+	leaf2 := crypto.Keccak256Hash([]byte("Merkle"))
+
+	s.Run("no leaves", func() {
+		root, siblings, err := CreateProofs(nil, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(s.pristine[1], root)
+
+		s.Equal(0, len(siblings))
+	})
+
+	s.Run("one leaf", func() {
+		leaves := []model.Hash{leaf1}
+
+		root, siblings, err := CreateProofs(leaves, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(
+			crypto.Keccak256Hash(leaf1[:], s.pristine[0][:]),
+			root,
+		)
+
+		s.Equal(len(leaves)*height, len(siblings))
+		s.Equal(s.pristine[0], siblings[0])
+
+		leafSiblings := siblings[0 : 0+height]
+		s.Equal(root, rootFromSiblings(leaf1, 0, leafSiblings))
+	})
+
+	s.Run("two leaves", func() {
+		leaves := []model.Hash{leaf1, leaf2}
+		leavesCopy := make([]model.Hash, len(leaves))
+		copy(leavesCopy, leaves)
+
+		root, siblings, err := CreateProofs(leavesCopy, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(
+			crypto.Keccak256Hash(leaf1[:], leaf2[:]),
+			root,
+		)
+
+		s.Equal(len(leaves)*height, len(siblings))
+		s.Equal(leaf2, siblings[0*height+0])
+		s.Equal(leaf1, siblings[1*height+0])
+
+		for idx := range leaves {
+			leafSiblings := siblings[idx*height : idx*height+height]
+			s.Equal(root, rootFromSiblings(leaves[idx], idx, leafSiblings))
+		}
+	})
+
+	s.Run("three leaves", func() {
+		leaves := make([]model.Hash, 3)
+
+		_, _, err := CreateProofs(leaves, 1)
+		s.Require().NotNil(err)
+
+		s.ErrorContains(err, "too many leaves")
+	})
+}
+
+func (s *CreateProofsSuite) TestHeightTwo() {
+	height := 2
+	leaf1 := crypto.Keccak256Hash([]byte("Merkle"))
+	leaf2 := crypto.Keccak256Hash([]byte("trees"))
+	leaf3 := crypto.Keccak256Hash([]byte("are"))
+	leaf4 := crypto.Keccak256Hash([]byte("cool"))
+
+	s.Run("no leaves", func() {
+		root, siblings, err := CreateProofs(nil, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(s.pristine[2], root)
+
+		s.Equal(0, len(siblings))
+	})
+
+	s.Run("one leaf", func() {
+		leaves := []model.Hash{leaf1}
+
+		root, siblings, err := CreateProofs(leaves, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(
+			crypto.Keccak256Hash(
+				crypto.Keccak256Hash(leaf1[:], s.pristine[0][:]).Bytes(),
+				s.pristine[1][:],
+			),
+			root,
+		)
+
+		s.Equal(len(leaves)*height, len(siblings))
+		s.Equal(s.pristine[0], siblings[0])
+
+		leafSiblings := siblings[0 : 0+height]
+		s.Equal(root, rootFromSiblings(leaf1, 0, leafSiblings))
+	})
+
+	s.Run("two leaves", func() {
+		leaves := []model.Hash{leaf1, leaf2}
+		leavesCopy := make([]model.Hash, len(leaves))
+		copy(leavesCopy, leaves)
+
+		root, siblings, err := CreateProofs(leavesCopy, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(
+			crypto.Keccak256Hash(
+				crypto.Keccak256Hash(leaf1[:], leaf2[:]).Bytes(),
+				s.pristine[1][:],
+			),
+			root,
+		)
+
+		s.Equal(len(leaves)*height, len(siblings))
+		s.Equal(leaf2, siblings[0*height])
+		s.Equal(leaf1, siblings[1*height])
+		s.Equal(s.pristine[1], siblings[0*height+1])
+		s.Equal(s.pristine[1], siblings[1*height+1])
+
+		for idx := range leaves {
+			leafSiblings := siblings[idx*height : idx*height+height]
+			s.Equal(root, rootFromSiblings(leaves[idx], idx, leafSiblings))
+		}
+	})
+
+	s.Run("three leaves", func() {
+		leaves := []model.Hash{leaf1, leaf2, leaf3}
+		leavesCopy := make([]model.Hash, len(leaves))
+		copy(leavesCopy, leaves)
+
+		root, siblings, err := CreateProofs(leavesCopy, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(
+			crypto.Keccak256Hash(
+				crypto.Keccak256Hash(leaf1[:], leaf2[:]).Bytes(),
+				crypto.Keccak256Hash(leaf3[:], s.pristine[0][:]).Bytes(),
+			),
+			root,
+		)
+
+		s.Equal(len(leaves)*height, len(siblings))
+		s.Equal(leaf2, siblings[0*height])
+		s.Equal(leaf1, siblings[1*height])
+		s.Equal(s.pristine[0], siblings[2*height])
+		s.Equal(crypto.Keccak256Hash(leaf3[:], s.pristine[0][:]), siblings[0*height+1])
+		s.Equal(crypto.Keccak256Hash(leaf3[:], s.pristine[0][:]), siblings[1*height+1])
+		s.Equal(crypto.Keccak256Hash(leaf1[:], leaf2[:]), siblings[2*height+1])
+
+		for idx := range leaves {
+			leafSiblings := siblings[idx*height : idx*height+height]
+			s.Equal(root, rootFromSiblings(leaves[idx], idx, leafSiblings))
+		}
+	})
+
+	s.Run("four leaves", func() {
+		leaves := []model.Hash{leaf1, leaf2, leaf3, leaf4}
+		leavesCopy := make([]model.Hash, len(leaves))
+		copy(leavesCopy, leaves)
+
+		root, siblings, err := CreateProofs(leavesCopy, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(
+			crypto.Keccak256Hash(
+				crypto.Keccak256Hash(leaf1[:], leaf2[:]).Bytes(),
+				crypto.Keccak256Hash(leaf3[:], leaf4[:]).Bytes(),
+			),
+			root,
+		)
+
+		s.Equal(len(leaves)*height, len(siblings))
+		s.Equal(leaf2, siblings[0*height])
+		s.Equal(leaf1, siblings[1*height])
+		s.Equal(leaf4, siblings[2*height])
+		s.Equal(leaf3, siblings[3*height])
+		s.Equal(crypto.Keccak256Hash(leaf3[:], leaf4[:]), siblings[0*height+1])
+		s.Equal(crypto.Keccak256Hash(leaf3[:], leaf4[:]), siblings[1*height+1])
+		s.Equal(crypto.Keccak256Hash(leaf1[:], leaf2[:]), siblings[2*height+1])
+		s.Equal(crypto.Keccak256Hash(leaf1[:], leaf2[:]), siblings[3*height+1])
+
+		for idx := range leaves {
+			leafSiblings := siblings[idx*height : idx*height+height]
+			s.Equal(root, rootFromSiblings(leaves[idx], idx, leafSiblings))
+		}
+	})
+
+	s.Run("five leaves", func() {
+		leaves := []model.Hash{leaf1, leaf2, leaf3, leaf4, leaf1}
+
+		_, _, err := CreateProofs(leaves, uint(height))
+		s.Require().NotNil(err)
+
+		s.ErrorContains(err, "too many leaves")
+	})
+}
+
+func (s *CreateProofsSuite) TestHeightThree() {
+	height := 3
+	leaf1 := crypto.Keccak256Hash([]byte("Merkle"))
+	leaf2 := crypto.Keccak256Hash([]byte("trees"))
+	leaf3 := crypto.Keccak256Hash([]byte("are"))
+	leaf4 := crypto.Keccak256Hash([]byte("so"))
+	leaf5 := crypto.Keccak256Hash([]byte("much"))
+	leaf6 := crypto.Keccak256Hash([]byte("fun"))
+	leaf7 := crypto.Keccak256Hash([]byte("wow"))
+	leaf8 := crypto.Keccak256Hash([]byte("!"))
+
+	s.Run("no leaves", func() {
+		root, siblings, err := CreateProofs(nil, uint(height))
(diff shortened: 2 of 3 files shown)


internal/merkle/tree.go
```diff
@@ -0,0 +1,178 @@
...
+func (t *Tree) node(level, index uint) Hash {
+	if level <= t.height && int(index) < len(t.nodes[level]) {
+		return t.nodes[level][index]
+	}
+	return t.pristine.nodes[level]
+}
+
+// maxLeaves calculates the maximum number of leaves for a given tree height.
+func maxLeaves(height uint) uint64 {
+	if height >= MAX_TREE_HEIGHT {
```
diegonehab, line 174: 2^64 does not fit in a uint64. So your max_tree_height has to be 63, right?

internal/merkle/tree.go
```diff
@@ -0,0 +1,178 @@
...
+	"errors"
+	"fmt"
+	"math"
+
+	. "github.com/cartesi/rollups-node/internal/node/model"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// MAX_TREE_HEIGHT defines the maximum allowable height for a Merkle tree.
+const MAX_TREE_HEIGHT = 64
```
diegonehab, line 16: See comment on `maxLeaves` below.

internal/merkle/tree.go
```diff
@@ -0,0 +1,178 @@
...
+func (t *Tree) RootHash() Hash {
+	return t.nodes[0][0]
+}
+
+// calculateRootHash recalculates the root hash of the tree.
+//
+// This method updates the hash values at each level of the tree
+// starting from the leaves up to the root. It uses the pristine tree nodes
+// to complete any levels that are not yet full.
+func (t *Tree) calculateRootHash() {
```
diegonehab, line 119: Does this work when the tree is empty?

internal/merkle/tree.go
```diff
@@ -0,0 +1,178 @@
...
+//
+// If the tree is already full it returns an error.
+func (t *Tree) Push(leaf Hash) error {
+	leafCount := len(t.nodes[t.height])
+	if t.height == 0 || uint64(leafCount) == maxLeaves(t.height) {
+		return errors.New("merkle: reached maximum capacity")
+	}
+
+	t.nodes[t.height] = append(t.nodes[t.height], leaf)
+	t.calculateRootHash()
```
diegonehab, line 89: This is *very* inefficient. When you append a new leaf, you only need to update one node per level, the "rightmost" node. Here, you are recomputing every internal node.

internal/merkle/tree.go
```diff
@@ -0,0 +1,178 @@
...
+// if the height exceeds MAX_TREE_HEIGHT.
+func NewTree(height uint) (*Tree, error) {
+	if height > MAX_TREE_HEIGHT {
+		return nil, fmt.Errorf("merkle: maximum tree height is %d", MAX_TREE_HEIGHT)
+	}
+	tree := new(Tree)
+	tree.height = height
+	tree.pristine = NewPristineTree(height)
+	tree.nodes = make([][]Hash, height+1) // the root is at height 0
+	tree.nodes[0] = make([]Hash, 1)       // sets the pristine node as the root
```
diegonehab, line 47: How is this root "pristine"? Shouldn't you copy it from the pristine tree?

internal/merkle/tree.go
```diff
@@ -0,0 +1,178 @@
...
+	if uint64(index) >= maxLeaves(t.height) {
+		return nil, errors.New("merkle: index out of bounds")
+	}
+	siblings := make([]Hash, 0, t.height)
+	for level := t.height; level > 0; level-- {
+		siblingIndex := index - 1
+		if index%2 == 0 {
+			siblingIndex = index + 1
+		}
+		siblings = append(siblings, t.node(level, siblingIndex))
```
diegonehab, line 156: Shouldn't you be dividing index by 2 at each iteration here?

diegonehab: And if so, how is it that this is passing the tests? Shouldn't the tests have caught this?

diegonehab (changes requested):
I think that when designing a datastructure like this, you have to keep in mind the use cases in the node.

I don't see a situation in which the node will have to keep this tree up-to-date as each leaf is being inserted. If there was such a use case, then please see comments in the code as to how to make this much more efficient. O(N D) instead, where N is the number of leaves and D the tree depth, instead of the current O(N^2) algorithm.

The use case I can see is a "batch" operation. The node will simply append the leaves as they come, and once the epoch is closed, it will need to produce everything. This can also be done in O(N D) without there even being a tree structure.

Let me know what use cases you had in mind.

guidanoli:
I think I agree with Diego on this one. You only need to generate the epoch hash at the end of every epoch, and this tree is brand new every epoch. This will, of course, be different after Output Unification v2, with an ever-growing output tree.

For now, I think this code can be a lot simpler. You might not even need to define a type for pristine trees. You can just calculate the node of the pristine tree at the current level on-the-fly when going from bottom to top.

torives:
Thanks, @diegonehab and @guidanoli for your review. I've removed the tree data structure and went with the batch strategy previously mentioned. I've also re-written the tests (inspired by the Solidity test code) and am more confident it is doing what it should. Please, take another look.

internal/merkle/proof.go
```diff
@@ -0,0 +1,100 @@
...
+	currentLevel []model.Hash,
+	siblings [][]model.Hash,
+	pristineNode model.Hash,
+) {
+	// for each leaf
+	for idx := range siblings {
+		// creates the sibling slice if needed
+		if siblings[idx] == nil {
+			siblings[idx] = make([]model.Hash, height)
+		}
```
guidanoli, line 65: Why not pre-allocate the whole matrix in `CreateProofs`?

internal/merkle/proof.go
```diff
@@ -0,0 +1,100 @@
...
+// If the level is not full, a pristine node is used instead.
+func calculateParents(currentLevel, nextLevel []model.Hash, pristineNode model.Hash) {
+	// for each parent node
+	for idx := range nextLevel {
+		leftChild := currentLevel[2*idx]
+		rightChild := pristineNode
+		// if current level is full
+		if 2*idx+1 < len(currentLevel) {
+			rightChild = currentLevel[2*idx+1]
+		}
```
guidanoli, line 96: Another use of the `at` function.

```suggestion
		rightChild := at(currentLevel, 2*idx+1, pristineNode)
```

internal/merkle/proof.go
```diff
@@ -0,0 +1,100 @@
...
+			sibling = currentLevel[siblingIdx]
+		}
+		siblings[idx][level] = sibling
+	}
+}
+
+// calculateParents computes the parent nodes for the next level in the Merkle
+// tree by hashing their children with the Keccak-256 hash function.
+//
+// If the level is not full, a pristine node is used instead.
```
guidanoli, line 87: When I think of a full level, I think it has $2^h$ non-pristine nodes. But it may be not full, but a pristine node is still used, if it has an odd number of non-pristine nodes.

internal/merkle/proof.go
```diff
@@ -0,0 +1,100 @@
...
+		return model.Hash{}, nil, errors.New("too many leaves for height")
+	}
+
+	return currentLevel[0], siblings, nil
+}
+
+// calculateSiblings iterates over each leaf and populates the siblings matrix
+// with the appropriate sibling nodes at the current level.
+//
+// If the level is not full, a pristine node is used instead.
```
guidanoli, line 53: I would also rephrase this to something like...

> If the current level has an odd number of nodes, a pristine node will be used.

internal/merkle/proof.go
```diff
@@ -0,0 +1,100 @@
...
+		nextLevel = aux[:(len(currentLevel)+1)/2]
+		pristineNode = crypto.Keccak256Hash(pristineNode[:], pristineNode[:])
+	}
+
+	// in the end, current level is the root level
+	if len(currentLevel) != 1 {
+		return model.Hash{}, nil, errors.New("too many leaves for height")
+	}
+
+	return currentLevel[0], siblings, nil
```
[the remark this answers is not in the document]
torives: The `at` function is quite elegant. I liked it very much when reading the Solidity code but forgot to implement it here as well 😅. Thanks.

internal/merkle/proof.go
```diff
@@ -0,0 +1,100 @@
...
+		leaves = append(leaves, pristineNode)
+	}
+
+	currentLevel := leaves
+	nextLevel := make([]model.Hash, (len(leaves)+1)/2)
+	siblings := make([][]model.Hash, len(leaves))
+
+	// for each level in the tree, starting from the leaves
+	for levelIdx := 0; levelIdx < int(height); levelIdx++ {
+		calculateSiblings(levelIdx, int(height), currentLevel, siblings, pristineNode)
```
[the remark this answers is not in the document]
torives: It was a workaround as Golang requires both operands to be the same type. For brevity, I preferred `height` cast to int instead of `levelIdx` to uint, but converting `calculateSiblings` parameters to unsigned ints is better 👍🏽

internal/merkle/proof.go
torives: Because I would have to loop over the leaves one more time just for this purpose

internal/merkle/proof.go
torives: You're correct, I mixed up the concepts here.

internal/merkle/proof.go
```diff
@@ -0,0 +1,100 @@
...
+			siblings[idx] = make([]model.Hash, height)
+		}
+
+		// the sibling index is to the left of the parent if leaf index is odd
+		// or to the right if leaf index is even
+		siblingIdx := (idx / 2) ^ 1
+		if level == 0 {
+			// in the leaf level, use the leaf index directly
+			siblingIdx = idx ^ 1
+		}
```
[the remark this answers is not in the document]
torives: I guess it could be if I were iterating over the nodes at the current level. But I'm always iterating over the leaves, as I found it easier to handle the matrix indexes.

internal/merkle/proof.go
guidanoli: That's true, but it would remove this `if` clause from your code.
On `CreateProofs`, you'd know it is `nil`.
On `CreateSiblings`, you'd like it is not `nil`.

internal/merkle/proof.go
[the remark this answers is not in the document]
guidanoli: I didn't know Golang didn't coerce 0 into `uint`. You could do this then...

```go
	for levelIdx := uint(0); levelIdx < height; levelIdx++ {
```

internal/merkle/proof.go
diegonehab: Why not use a single allocation? I.e., allocate an array with `len(leaves)*height` entries, and then access it with index (i,j) using i*height+j, where i is the leaf j is the level in the tree. Or you can transpose it, if you want, and access it as j*len(leaves)+i. Doesn't matter if you use the so-called column-major or row-major order. Either way it's a single allocation. It's how people represent matrices in "real life".

internal/merkle/proof.go
```diff
@@ -0,0 +1,100 @@
...
+	currentLevel := leaves
+	nextLevel := make([]model.Hash, (len(leaves)+1)/2)
+	siblings := make([][]model.Hash, len(leaves))
+
+	// for each level in the tree, starting from the leaves
+	for levelIdx := 0; levelIdx < int(height); levelIdx++ {
+		calculateSiblings(levelIdx, int(height), currentLevel, siblings, pristineNode)
+		calculateParents(currentLevel, nextLevel, pristineNode)
+
+		aux := currentLevel
```
diegonehab, line 32: Come on, you are using a language that has multi-value assignments. Embrace it and write 

```golang
currentLevel, nextLevel = nextLevel, currentLevel[:(len(currentLevel)+1)/2]
```

internal/merkle/proof.go
```diff
@@ -0,0 +1,95 @@
...
+func calculateParents(currentLevel, nextLevel []model.Hash, pristineNode model.Hash) {
+	// for each parent node
+	for idx := range nextLevel {
+		leftChild := currentLevel[2*idx]
+		rightChild := at(currentLevel, uint(2*idx+1), pristineNode)
+		nextLevel[idx] = crypto.Keccak256Hash(leftChild[:], rightChild[:])
+	}
+}
+
+func at(array []model.Hash, index uint, defaultValue model.Hash) model.Hash {
```
diegonehab, line 89: Add documentation?

internal/merkle/proof.go
```diff
@@ -0,0 +1,95 @@
...
+
+		siblings[idx][levelIdx] = at(currentLevel, uint(siblingIdx), pristineNode)
+	}
+}
+
+// calculateParents computes the parent nodes for the next level in the Merkle
+// tree by hashing their children with the Keccak-256 hash function.
+//
+// If the current level has an odd number of nodes a pristine node will be used.
+func calculateParents(currentLevel, nextLevel []model.Hash, pristineNode model.Hash) {
```
diegonehab, line 80: I think this is a matter of taste, but I see an error prone pattern here. The input is `currentLevel`. `nextLevel` is the output. And yet, you are reading the size of the *output* to figure out how many times to loop. I think it would be best to reslice `nextLevel` inside this function, rather than on the outside. Right?

diegonehab: Is there any convention in Go regarding where inputs and outputs go in the order of arguments in a function? In C++, we typically do inputs-first-then-outputs.

internal/merkle/proof.go
```diff
@@ -0,0 +1,95 @@
...
+	// for each parent node
+	for idx := range nextLevel {
+		leftChild := currentLevel[2*idx]
```
diegonehab, line 83: I *think* both these definitions create copies, right? Shouldn't you use `[:]` to avoid that? Or am I mistaken?

diegonehab: I see this pattern elsewhere in the code. Or is it copy-on-write? Do you know?

internal/merkle/proof.go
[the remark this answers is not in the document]
diegonehab: I don't get it either. Can you explain? This doesn't seem right at all when level is `> 1`...

diegonehab (dismissed):
Please see comments. I am worried that, once again, your tests are not comprehensive enough. Unless the potential bug @guidanoli and I identified by code inspection are really not bugs.

guidanoli:
> your tests are not comprehensive enough.

I agree. One nice approach is to do random tests. Pick a random $n \in [1,N]$ for a generous constant $N$ (say, 100). Then, pick $n$ random hashes $h_i$ for $i \in [1,n]$. Then, construct a Merkle tree with these $n$ hashes as leaves. Construct proofs for each of them. Then, verify each proof. Also, you can pick a random sibling in the proof and replace it by a different hash, and check that the proof doesn't match, just to catch false positives in the verification algorithm.

internal/merkle/proof.go
guidanoli: You could even write the next level in-place, given that you've already extracted the siblings at the current level with `calculateSiblings`. Therefore, this function could have the following signature:

```go
func calculateParents(level []model.Hash, pristineNode model.Hash)
```

I don't know if it is possible to slice an array in-place in Golang. If not, this function could return the size of the slice, so that the caller does the slicing. With this, the `CreateProofs` function wouldn't even need to maintain a `nextLevel`, just the `currentLevel`.

internal/merkle/proof.go
guidanoli: That's true. It seems that the `model.Hash` type is, in fact, just an array of 32 bytes. So, maybe `[:]` is necessary to avoid copying hashes around? I don't know much about Golang.

internal/merkle/proof.go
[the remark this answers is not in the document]
torives: You are correct, I re-checked my math and realized the error. I'll fix the code and enhance the test suite as suggested by @guidanoli.

internal/merkle/proof.go
diegonehab: I think using two levels is fine. Diminishing returns.

diegonehab:
The thing is that you should try to cover the basic cases. You haven’t tested the sibling hashes by bubbling up from leaf to root to see if it matches the root hash. You should take a small height (say, 3 or 4) and test *everything*. All number of leaves. The siblings for all of them. Too many leaves. Zero leaves. Ideally, you should generate the expected results using some other code that we already have either in solidity, Lua, or C++.

internal/merkle/proof.go
torives: > I _think_ both these definitions create copies, right?

Yes, they're both copies.

> Shouldn't you use `[:]` to avoid that? Or am I mistaken?

I could. But since I need to modify the `at` function to use pointers for the same reason (avoiding copies), I would rather use pointers directly.

diegonehab: Don't know much about Golang, but why pointers? I realize there is no concept of a const slice (go figure...).

internal/merkle/proof.go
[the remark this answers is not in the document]
torives: Done at a764662892a930eda44dd48b1a1278f0da722e41

internal/merkle/proof.go
torives: Done at a764662892a930eda44dd48b1a1278f0da722e41

internal/merkle/proof.go
torives: I liked @guidanoli suggestion and ended up implementing it at 293e57080aa44d2aded47e53f6ef202c4c99295b.

internal/merkle/proof.go
torives: This code changed a bit because of [this](https://github.com/cartesi/rollups-node/pull/444#discussion_r1679388012). Changes can be found at 293e57080aa44d2aded47e53f6ef202c4c99295b.

internal/merkle/proof.go
torives: > I realize there is no concept of a const slice (go figure...).
AFAIK in Golang arrays are meant to be fixed and slices are meant to be dynamic. That is just the way it is 🤷🏽‍♂️ 

> Don't know much about Golang, but why pointers?
It is the only way we can avoid copies In the current situation (a slice of byte arrays). It didn't look so bad in the end. Done at a66d705dc0b807351ce69d39378a031e3df117e3.

internal/merkle/proof_test.go
```diff
@@ -247,3 +971,86 @@ func (s *ComputeProofSuite) TestItMatchesMachineImplementation() {
 		root,
 	)
 }
+
+func FuzzVerifyProofs(f *testing.F) {
+	f.Add(uint(10), uint(10), uint(5))
+	f.Fuzz(func(t *testing.T, height, leafCount, leafIdx uint) {
+		height = bound(height, 3, 20)
+		leafCount = bound(leafCount, 1, uint(math.Pow(2, float64(height))))
```
guidanoli, line 979: I think we can use bit shifting here.

```suggestion
		leafCount = bound(leafCount, 1, 1 << height)
```

internal/merkle/proof_test.go
```diff
@@ -247,3 +971,86 @@ func (s *ComputeProofSuite) TestItMatchesMachineImplementation() {
...
+		if root != newRoot {
+			t.Errorf("expected: %v, got: %v\n", root, newRoot)
+		}
+
+		newRoot = rootFromSiblings(model.Hash{}, int(leafIdx), leafSiblings)
+		if root == newRoot {
+			t.Errorf("expected root to be different when replacing the leaf")
+		}
+
+		leafSiblings[0] = crypto.Keccak256Hash([]byte("wrong_sibling"))
```
guidanoli, line 1001: You could select a random sibling as well.

internal/merkle/proof_test.go
```diff
@@ -247,3 +971,86 @@ func (s *ComputeProofSuite) TestItMatchesMachineImplementation() {
...
+			t.Error(err)
+		}
+
+		leafSiblings := siblings[leafIdx*height : leafIdx*height+height]
+		newRoot := rootFromSiblings(leaves[leafIdx], int(leafIdx), leafSiblings)
+		if root != newRoot {
+			t.Errorf("expected: %v, got: %v\n", root, newRoot)
+		}
+
+		newRoot = rootFromSiblings(model.Hash{}, int(leafIdx), leafSiblings)
```
guidanoli, line 996: I would check that the leaf is non-zero just to be safe.

guidanoli (commented):
Looking good!

internal/merkle/proof.go
diegonehab: When you assign a reslice of an array to a variable, you don’t create a copy, unless I’m crazy.

internal/merkle/proof_test.go
```diff
@@ -61,61 +66,110 @@ func (s *ComputeProofSuite) TestHeightOne() {
...
+		s.Equal(s.pristine, siblings[0])
+
+		leafSiblings := siblings[0 : 0+height]
+		s.Equal(root, rootFromSiblings(leaf1, 0, leafSiblings))
+	})
+
 	s.Run("two leaves", func() {
 		leaves := []model.Hash{leaf1, leaf2}
+		leavesCopy := make([]model.Hash, len(leaves))
+		copy(leavesCopy, leaves)
```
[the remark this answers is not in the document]
torives: So I can grab a leaf by its index later when verifying the proofs. Since `CreateProofs` uses `leaves`' underlying array to do its thing, I can't reuse it.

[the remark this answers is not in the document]
guidanoli: I see, so we leave it up to the caller to decide whether they want to copy the array of leaves before calling `CreateProofs`.

internal/merkle/proof_test.go
torives: 🤯 🔫

internal/merkle/proof_test.go
```diff
@@ -247,3 +971,86 @@ func (s *ComputeProofSuite) TestItMatchesMachineImplementation() {
...
+func bound(n, min, max uint) uint {
+	if max < min {
+		panic("max should be equal or greater than min")
+	}
+	if n < min {
+		return min
+	} else if n > max {
+		return max
+	}
+	return n
```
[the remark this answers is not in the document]
torives: Done at 35c518c6956338e4a67982cadf920ddd7942f128

internal/merkle/proof_test.go
```diff
@@ -61,61 +66,110 @@ func (s *ComputeProofSuite) TestHeightOne() {
...
 
+	s.Run("no leaves", func() {
+		root, siblings, err := CreateProofs(nil, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(
+			crypto.Keccak256Hash(
+				crypto.Keccak256Hash(s.pristine[:], s.pristine[:]).Bytes(),
+				crypto.Keccak256Hash(s.pristine[:], s.pristine[:]).Bytes(),
+			),
```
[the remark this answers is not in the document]
torives: Done at 6fc865150288bda76d2a6ebd46d3cc1b34935f8a

internal/merkle/proof_test.go
```diff
@@ -61,61 +66,110 @@ func (s *ComputeProofSuite) TestHeightOne() {
...
+	s.Run("one leaf", func() {
+		leaves := []model.Hash{leaf1}
+
+		root, siblings, err := CreateProofs(leaves, uint(height))
+		s.Require().Nil(err)
+
+		s.Equal(
+			crypto.Keccak256Hash(
+				crypto.Keccak256Hash(leaf1[:], s.pristine[:]).Bytes(),
+				crypto.Keccak256Hash(s.pristine[:], s.pristine[:]).Bytes(),
```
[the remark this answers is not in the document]
torives: Done at 6fc865150288bda76d2a6ebd46d3cc1b34935f8a

internal/merkle/proof_test.go
torives: Done at 26708fc3d1d198807daf41d1303e070ca5048a90

internal/merkle/proof_test.go
torives: Done at 92c00d6ec1ebed0c913cffac657d70596da188d7.

internal/merkle/proof_test.go
torives: Done at ad110882c5bdab743286d0c6dc25582b07587f32

guidanoli (dismissed):
Will re-approve if needed after squash.

internal/merkle/proof.go
torives: My previous comment wasn't completely true. I could also re-slice it to avoid a copy as you originally proposed, but then I would end up with a `[]byte` slice, instead of a `model.Hash`. This would defeat the purpose of having a custom type to represent hashes so I went with a pointer to `model.Hash` instead.

diegonehab: To be fair, that is exactly what the implementation of `crypto.Keccak256Hash`. This seems to be idiomatic, no? I am not a Go specialist.

torives: I agree with you. Just because I needed to use pointers for the `at` function it doesn't mean I have to use them here also when there's a more idiomatic way. I changed it at 9eb81fc6e7b4c11d24a4649bf7823edaa1c6c05a.

internal/merkle/proof.go
```diff
@@ -0,0 +1,78 @@
...
+// with the given height. It returns the root hash and the siblings matrix
+// encoded as an array, in bottom-up order.
+//
+// If the number of leaves exceeds the capacity for the given height,
+// an error is returned.
+func CreateProofs(leaves []model.Hash, height uint) (model.Hash, []model.Hash, error) {
+	pristineNode := model.Hash{}
+
+	currentLevel := leaves
+	leafCount := uint(len(leaves))
```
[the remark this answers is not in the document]
torives: > The given height could be checked against the number of leaves and return the error `too many leaves for height`. 

We could, but that would force me to define and check the maximum height, adding a second `if` statement and error. I'd rather keep it as it is, looking as close as possible to the implementation we have in the [contracts](https://github.com/cartesi/rollups-contracts/blob/main/contracts/library/LibMerkle32.sol#L19).

internal/merkle/proof.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+		// calculate the next level
+		currentLevel = parentLevel(currentLevel, &pristineNode)
+		// update the pristine node for the next level
+		pristineNode = crypto.Keccak256Hash(pristineNode[:], pristineNode[:])
+	}
+
+	// in the end, current level is the root level
+	if len(currentLevel) > 1 {
+		return model.Hash{}, nil, errors.New("too many leaves for height")
```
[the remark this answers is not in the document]
torives: Done at a422f86e1510ec873cad8979e3f9e8a7d4999c4f

internal/merkle/proof_test.go
```diff
@@ -0,0 +1,877 @@
...
+	root, _, err := CreateProofs(leaves, 16)
+	s.Require().Nil(err)
+
+	s.Equal(
+		common.HexToHash("0xe8e0477114cb630c4d14eea249eb2c63d84c9c685ddf35d137019e659ae20418"),
+		root,
+	)
+}
+
+func FuzzVerifyProofs(f *testing.F) {
```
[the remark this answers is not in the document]
torives: I'm not sure I understood your goal. All fuzzy tests are executed along the unit tests, using the data added programatically or located in the `testdata` directory if that's what you meant.

internal/merkle/proof_test.go
```diff
@@ -0,0 +1,877 @@
...
+		s.Equal(leaf2, siblings[0*height+0])
+		s.Equal(leaf1, siblings[1*height+0])
+
+		for idx := range leaves {
+			leafSiblings := siblings[idx*height : idx*height+height]
+			s.Equal(root, rootFromSiblings(leaves[idx], idx, leafSiblings))
+		}
+	})
+
+	s.Run("three leafs", func() {
```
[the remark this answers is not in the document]
torives: Done in 09efe7b9a0e8a783ff101257114ee0c00f88d261

internal/merkle/proof_test.go
```diff
@@ -0,0 +1,877 @@
...
+		leaf := crypto.Keccak256Hash([]byte("Cartesi"))
+
+		root, siblings, err := CreateProofs([]model.Hash{leaf}, 0)
+		s.Require().Nil(err)
+
+		s.Equal(leaf, root)
+		s.Equal(0, len(siblings))
+	})
+
+	s.Run("two leafs", func() {
```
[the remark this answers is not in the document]
torives: Done in 09efe7b9a0e8a783ff101257114ee0c00f88d261.

.golangci.yml
```diff
@@ -12,5 +12,5 @@ linters-settings:
     line-length: 100
     tab-width: 4
   mnd:
-    ignored-functions:
-      - "^make"
+    ignored-numbers:
```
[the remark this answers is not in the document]
torives: Actually, I don't this is needed anymore. I removed it in [1a3371e](https://github.com/cartesi/rollups-node/pull/444/commits/1a3371e52ebba4e925d8722401910e39dcad6c18).

fmoura (dismissed):

marcelstanley (approved):

fmoura (dismissed):

internal/merkle/proof.go
```diff
@@ -0,0 +1,79 @@
...
+	leafCount := uint(len(leaves))
+	siblings := make([]model.Hash, leafCount*height)
+
+	// for each level in the tree, starting from the leaves
+	for levelIdx := uint(0); levelIdx < height; levelIdx++ {
+		// for each leaf
+		for leafIdx := uint(0); leafIdx < leafCount; leafIdx++ {
+			// calculate its sibling at the current level.
+			// The sibling index is to the left of the parent if leaf index
+			// is odd or to the right if even
```
guidanoli, line 32: I would expand on this. Feel free to adapt it however you like.

```suggestion
			// Each pair of siblings shares the same parent.
			// For any given parent with index i,
			// the index of its left node is 2*i,
			// and the index of its right node is 2*i+1.
			// Since their indices differ only by the LSB,
			// by flipping the LSB with ^1,
			// we calculate the index of one another.
			// In order to get the index of a parent,
			// you remove the LSB with >>1.
			// For an ancestor of height difference h,
			// you remove the h LSBs with >>h.
```

marcelstanley (approved):

fmoura (approved):

PR #445, opened 2024-06-13
fix!: close epochs based on block numbers 

### LINKED ISSUE #432: [v.1.4.0] ClaimMismatch on authority-claimer
As reported by a community member, the authority-claimer:

> doesn't seem to recover when running a dapp with `cartesi cli version @cartesi/cli/0.14.0 linux-x64 node-v21.7.2`".

## 🙂 Expected behavior

- Claims should be sent at the end of every epoch and duplicate ones should be discarded. 
- The `authority claimer` should keep working.

## 🫠 Actual behavior

Error `DuplicatedClaimError { source: ClaimMismatch { expected_first_index: 37, claim_first_index: 36, claim_last_index: 39 } } service=authority-claimer` occurs and the authority-claimer exits, bringing the node to a full stop.

Full logs:
[compassionate_brattain_logs-ONLY_FIRST_RUN.txt](https://github.com/cartesi/rollups-node/files/15485661/compassionate_brattain_logs-ONLY_FIRST_RUN.txt)

## 🧪 Minimal test case

No test case available so far.

## 🌎 Environment

- @cartesi/cli/0.14.0 linux-x64 node-v21.7.2 (Rollups Node 1.4.0)
- Sepolia network

renan061:


marcelstanley (commented):
A few suggestions after a first round:

marcelstanley:
BTW, @renan061, I'd suggest changing the commit message to something like 

```
fix: close epochs based on block numbers
```

And then, give more details in a few additionallines

CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+### Changed
+
+- Changed the dispatcher to close epochs based on block numbers instead of block timestamps.
+- Renamed `CARTESI_EPOCH_DURATION` to `CARTESI_EPOCH_LENGTH`, and set its default value to 5760 (1 day worth of blocks, in average).
```
marcelstanley, line 16: Please use `CARTESI_EPOCH_BLOCK_LENGTH` instead.

```suggestion
- **BREAKING**: replaced `CARTESI_EPOCH_DURATION` with `CARTESI_EPOCH_BLOCK_LENGTH` to match the new epoch calculation procedure, and set its default value to 5760 (1 day worth of blocks, in average).
```

torives: I also feel there should be a reference to the `ClaimMismatch` error being fixed. Something like:

```
### Fixed
- Fixed a bug that caused the node to create a ClaimMismatch error when reprocessing inputs after a reboot.
```

torives: I also feel there should be a reference to the `ClaimMismatch` bug being fixed. Something like:

```
### Fixed
- Fixed a bug that caused a `ClaimMismatch` error when reprocessing inputs after a reboot.
```

offchain/dispatcher/src/drivers/context.rs
```diff
@@ -11,48 +11,53 @@ use types::foldables::Input;
...
+        genesis_block: u64,
         epoch_length: u64,
         dapp_metadata: DAppMetadata,
         metrics: DispatcherMetrics,
-        status: RollupStatus,
+        inputs_sent: u64,
+        last_input_epoch: Option<u64>,
+        last_finished_epoch: Option<u64>,
     ) -> Self {
+        assert!(epoch_length > 0);
```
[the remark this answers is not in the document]
renan061: I added one assertion, to `should_finish_epoch`:
```
        match (self.last_input_epoch, self.last_finished_epoch) {
            (Some(input), Some(finished)) => assert!(input >= finished),
            _ => (),
        }
```

offchain/dispatcher/src/drivers/mock.rs
```diff
@@ -77,16 +77,16 @@ pub fn update_input_box(
 // ------------------------------------------------------------------------------------------------
 
 #[derive(Debug, Clone, Copy, PartialEq)]
-pub enum SendInteraction {
-    EnqueuedInput(u64),
-    FinishedEpoch(u64),
+pub enum Sent {
```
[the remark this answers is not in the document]
renan061: Went with `Event`.

renan061:
> BTW, @renan061, I'd suggest changing the commit message to something like
> 
> ```
> fix: close epochs based on block numbers
> ```
> 
> And then, give more details in a few additional lines

Changed to:

```
fix!: close epochs based on block numbers

- Fixes the ClaimMismatch bug on v.1x.
- BREAKING CHANGE: replaces the CARTESI_EPOCH_DURATION environment
  variable with CARTESI_EPOCH_LENGTH_IN_BLOCKS.
```

offchain/dispatcher/src/drivers/context.rs
```diff
@@ -62,98 +67,122 @@ impl Context {
...
-            let current_epoch = self.calculate_epoch(self.last_timestamp);
-            let event_epoch = self.calculate_epoch(event_timestamp);
-            event_epoch > current_epoch
+    fn calculate_epoch(&self, block: u64) -> u64 {
+        assert!(block >= self.genesis_block);
+        (block - self.genesis_block) / self.epoch_length
+    }
+
+    fn should_finish_epoch(&self, epoch: u64) -> bool {
+        match (self.last_input_epoch, self.last_finished_epoch) {
```
vfusco, line 95: Shouldn't this assert be elsewhere? eg.: When writing this values?

offchain/dispatcher/src/drivers/context.rs
```diff
@@ -62,98 +67,122 @@ impl Context {
...
+            (Some(input), Some(finished)) => assert!(input >= finished),
+            _ => (),
+        }
+
+        if self.last_finished_epoch == self.last_input_epoch {
```
vfusco, line 100: This conditions should be better explained in a comment, on why they work.

vfusco (changes requested):

fmoura (dismissed):
/setup_env.sh and /build/compose-devnet.yaml must be updated with `CARTESI_EPOCH_LENGTH_IN_BLOCKS` replacing `CARTESI_EPOCH_DURATION`

test/config.go
```diff
@@ -22,7 +20,7 @@ const (
 	LocalHttpPort                         = 10000
 	LocalBlockTimeout                     = 120
 	LocalFinalityOffset                   = 1
-	LocalEpochDurationInSeconds           = 240
+	LocalEpochDurationInSeconds           = 20
```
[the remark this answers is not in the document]
fmoura: Added some more fix suggestions at https://github.com/cartesi/rollups-node/pull/453

offchain/dispatcher/src/drivers/context.rs
renan061: Changed to:
```      
// If both are None, then there are no inputs by definition, and the epoch is empty.
// Otherwise, at least one epoch was finished, and the last enqueued input belongs to that
// epoch; meaning that any subsequent epochs (which includes the current one) are empty.
if self.last_finished_epoch == self.last_input_epoch {
        return false; // if the current epoch is empty
}
```

offchain/dispatcher/src/drivers/context.rs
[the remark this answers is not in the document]
renan061: Moved this assertion to the `enqueue_input` function.

offchain/dispatcher/src/drivers/context.rs
```diff
@@ -62,98 +67,122 @@ impl Context {
         input: &Input,
         broker: &impl BrokerSend,
     ) -> Result<(), BrokerFacadeError> {
-        broker.enqueue_input(self.inputs_sent_count, input).await?;
+        let input_block = input.block_added.number.as_u64();
+        self.finish_epoch_if_needed(input_block, broker).await?;
```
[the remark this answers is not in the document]
renan061: The other one reacts to a random block at the end of all the enqueues.
It is doing something conceptually different.

The deal with `finish_epoch_if_needed` inside the enqueue has to do with the consistency of the `Context`.
Imagine if we didn't have it inside the enqueue. Now let's look at the following order of events:
- enqueue(input from block 0)
- enqueue(input from block 1)
- enqueue(input from block 2)

Given that the epoch length is 2, and genesis  is 0, this would create 3 consecutive inputs without closing the epoch in between, which taints `last_input_epoch` and `last_finished_epoch`.

Thus, because this relates to the internal consistency of the `Context` module, it should belong to the `Context` module.
Before, this logic was divided between `context.rs` and `machine.rs`, which makes it difficult to reason with (and even test).

offchain/dispatcher/src/drivers/context.rs
renan061: Also created this assertion in `finish_epoch`:
```
assert!(self.last_input_epoch.is_some());
self.last_finished_epoch = self.last_input_epoch;
```

offchain/dispatcher/src/drivers/context.rs
renan061: Added:
```
// If both are None, then there are no inputs by definition, and the current epoch is empty.
//
// If both are Some(_), then at least one epoch was finished, and the last enqueued
// input belongs to that epoch; meaning that any subsequent epochs (which includes the
// current one) are empty.
//
// If last_finished_epoch is Some(_) and last_input_epoch is None, then we have an
// inconsistent internal state. The assertion from `finish_epoch` prevents this case from
// ever happening.
//
// If last_finished_epoch is None and last_input_epoch is Some(_), then the epoch is not
// empty by definition.
if self.last_finished_epoch == self.last_input_epoch {
    return false; // if the current epoch is empty
}
```

renan061: Actually, I'm changing this. WIP. brb.

internal/node/services.go
```diff
@@ -160,8 +160,7 @@ func newDispatcher(c config.NodeConfig, workDir string) services.CommandService
 	s.Env = append(s.Env, fmt.Sprintf("HISTORY_ADDRESS=%v", c.ContractsHistoryAddress))
 	s.Env = append(s.Env, fmt.Sprintf("AUTHORITY_ADDRESS=%v", c.ContractsAuthorityAddress))
 	s.Env = append(s.Env, fmt.Sprintf("INPUT_BOX_ADDRESS=%v", c.ContractsInputBoxAddress))
-	s.Env = append(s.Env, fmt.Sprintf("RD_EPOCH_DURATION=%v",
-		int(c.RollupsEpochDuration.Seconds())))
+	s.Env = append(s.Env, fmt.Sprintf("RD_EPOCH_DURATION=%v", c.RollupsEpochLengthInBlocks))
```
vfusco, line 163: RD_EPOCH_DURATION -> RD_EPOCH_LENGTH

What RD stands for?

vfusco (changes requested):

CHANGELOG.md
```diff
@@ -12,6 +12,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Fixed
+
+- Fixed a bug that caused a `authority_claimer::ClaimMismatch` error when reprocessing inputs after a reboot.
+
+### Changed
+
+- Changed the dispatcher to close epochs based on block numbers instead of block timestamps.
+- **BREAKING**: replaced the environment variable `CARTESI_EPOCH_DURATION` with `CARTESI_EPOCH_LENGTH_IN_BLOCKS` to match the new epoch algorithm, and set its default value to 7200 (1 day worth of blocks, in average, considering one block is mined every 12 seconds).
+- **BREAKING**: replaced the internal environment variable `RD_EPOCH_DURATION` with `RD_EPOCH_LENGTH_IN_BLOCKS` to match the new epoch algorithm, and also set its default value to 7200.
```
marcelstanley, line 23: Being an internal configuration, this would not require a changelog entry. I believe we overdid it here.

build/docker-bake.hcl
```diff
@@ -18,7 +18,7 @@ target "common" {
   context    = ".."
   args       = {
     BASE_IMAGE                = "debian:bookworm-20240311-slim"
-    RUST_VERSION              = "1.77.0"
+    RUST_VERSION              = "1.78.0"
```
marcelstanley, line 21: Please add an entry to the CHANGELOG for this bump.

offchain/host-runner/src/config.rs
```diff
@@ -5,6 +5,7 @@ use clap::Parser;
 use log::{LogConfig, LogEnvCliConfig};
 
 const DEFAULT_ADDRESS: &str = "0.0.0.0";
+
```
marcelstanley, line 8:
```suggestion
```

build/docker-bake.hcl
marcelstanley: In fact, I believe we should bump it to 1.79

CHANGELOG.md
```diff
@@ -12,6 +12,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Added Rollups end-to-end test using Echo Dapp
 
+### Fixed
+
+- Fixed a bug that caused a `authority_claimer::ClaimMismatch` error when reprocessing inputs after a reboot.
+
+### Changed
+
+- Changed the dispatcher to close epochs based on block numbers instead of block timestamps.
+- **BREAKING**: replaced the environment variable `CARTESI_EPOCH_DURATION` with `CARTESI_EPOCH_LENGTH_IN_BLOCKS` to match the new epoch algorithm, and set its default value to 7200 (1 day worth of blocks, in average, considering one block is mined every 12 seconds).
```
marcelstanley, line 22:
```suggestion
- **BREAKING**: replaced the environment variable `CARTESI_EPOCH_DURATION` with `CARTESI_EPOCH_LENGTH_IN_BLOCKS` to match the new epoch algorithm, and set its default value to 7200 (1 day worth of blocks, considering one block is mined every 12 seconds).
```

offchain/dispatcher/src/drivers/context.rs
```diff
@@ -62,98 +64,133 @@ impl Context {
...
-    // timestamp of each individual input, rather than just the latest from the blockchain.
-    fn should_finish_epoch(&self, event_timestamp: u64) -> bool {
-        if self.inputs_sent_count == 0 || self.last_event_is_finish_epoch {
-            false
-        } else {
-            let current_epoch = self.calculate_epoch(self.last_timestamp);
-            let event_epoch = self.calculate_epoch(event_timestamp);
-            event_epoch > current_epoch
+    fn calculate_epoch(&self, block_number: u64) -> u64 {
+        assert!(block_number >= self.genesis_block);
```
marcelstanley, line 90: This assert uncovers an issue on the `dispatcher` configuration, which uses the `application_deployment_block_number` instead of the `input_box_block_number` to check for valid inputs (`block_number`). This can be seen at https://github.com/cartesi/rollups-node/blob/70c5816ac155189f40e414a2df22d54d005b052d/offchain/dispatcher/src/setup.rs#L83.

We should replace that with the right configuration otherwise inputs sent before the application has been deployed will be considered as invalid by the assertion here.

marcelstanley (changes requested):

setup_env.sh
```diff
@@ -3,15 +3,14 @@ export CARTESI_LOG_PRETTY="true"
 export CARTESI_FEATURE_HOST_MODE="false"
 export CARTESI_FEATURE_DISABLE_CLAIMER="false"
 export CARTESI_FEATURE_DISABLE_MACHINE_HASH_CHECK="false"
-export CARTESI_EPOCH_DURATION="120"
+export CARTESI_EPOCH_LENGTH_IN_BLOCKS="10"
```
marcelstanley, line 6:
```suggestion
export CARTESI_EPOCH_LENGTH="10"
```

CHANGELOG.md
```diff
@@ -12,6 +12,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Fixed
+
+- Fixed a bug that caused a `authority_claimer::ClaimMismatch` error when reprocessing inputs after a reboot.
+
+### Changed
+
+- Changed the dispatcher to close epochs based on block numbers instead of block timestamps.
+- **BREAKING**: changed the dispatcher to use the `CARTESI_CONTRACTS_INPUT_BOX_DEPLOYMENT_BLOCK_NUMBER` environment variable instead of `CARTESI_CONTRACTS_APPLICATION_DEPLOYMENT_BLOCK_NUMBER`.
+- **BREAKING**: replaced the environment variable `CARTESI_EPOCH_DURATION` with `CARTESI_EPOCH_LENGTH` to match the new epoch algorithm, and set its default value to 7200 (1 day worth of blocks, in average, considering one block is mined every 12 seconds).
```
marcelstanley, line 23:
```suggestion
- **BREAKING**: replaced the environment variable `CARTESI_EPOCH_DURATION` with `CARTESI_EPOCH_LENGTH` to match the new epoch algorithm, and set its default value to 7200 (1 day worth of blocks, considering one block is mined every 12 seconds).
```

marcelstanley (dismissed):

offchain/dispatcher/src/drivers/context.rs
```diff
@@ -62,98 +64,144 @@ impl Context {
...
+        broker.finish_epoch(self.inputs_sent).await?;
         self.metrics
             .finish_epochs_sent
             .get_or_create(&self.dapp_metadata)
             .inc();
-        self.last_timestamp = event_timestamp;
-        self.last_event_is_finish_epoch = true;
+
+        // Asserting that there are inputs in the current epoch.
+        assert!(self.inputs_sent > 0);
```
vfusco, line 137: The asserts should come before `broker.finish_epoch`

vfusco (changes requested):

vfusco (dismissed):

vfusco (approved):

marcelstanley (approved):

Issue #368, opened 2024-04-03
Create PostGraphile view in the database


GMKrieger:
## 📚 Context

The GraphQL API we're going to expose to the public, more importantly the convenience team, can't be the whole database we're using to build the Node. Much of the information is for internal use, and the info needed by the clients is very specific, and need to be presented in a specific way.

As such, the tables and relations PostGraphile will use to generate the API need to be different from the core database.

## ✔️ Solution

PostGraphile generates a GraphQL API based on a single schema on the database, with all information contained in such schema. It's also able to generate the API from views, not necessarily tables.

With this, it's possible to create a schema with views derived from queries that collect only the information we wish to share from the database for the PostGraphile to generate the API from. This also insulates the GraphQL from most changes we wish to make on the database, and cuts down on refactoring or need for alignment with other teams.

## 📈 Subtasks

- [x] Create new migration for PostGraphile
- [x] Add PostGraphile startup to the node

PR #452, opened 2024-06-19
Add postgraphile view to database

GMKrieger:
#368


GMKrieger:
With the multi-dapp update to the database, this PR has become obsolete. I will close it now and it will be addressed on #478

PR #455, opened 2024-06-21
add --verbose option to cli send command

fmoura:


pkg/ethutil/ethutil.go
```diff
@@ -106,6 +108,12 @@ func getInputIndex(
 		if err != nil {
 			return 0, fmt.Errorf("failed to parse input added event: %v", err)
 		}
+
+		inputAddedJsonBytes, err := json.Marshal(inputAdded)
```
marcelstanley, line 112: Shouldn't it fail when `err != nil`?

marcelstanley:
BTW, by adding support to `slog`, the next logical step would be converting all logs to `slog`

pkg/ethutil/ethutil.go
fmoura: I had mixed feelings about this. I will do a different log , but not fail, as this deserialization is only used for debug logging

fmoura: Removed the Json Marshal. It was only useful for a very specific need.

pkg/ethutil/ethutil.go
```diff
@@ -106,6 +107,9 @@ func getInputIndex(
 		if err != nil {
 			return 0, fmt.Errorf("failed to parse input added event: %v", err)
 		}
+
+		slog.Debug("Input added", "event", inputAdded)
+
```
marcelstanley, line 112: Looking at this again, I think we should stay away from the ethutil package if the purpose of the change is simply logging the inputAdded event. 

I believe a better option would be simply getting the input index on the cli and using [`ethutil.GetInputFromInputBox()`](https://github.com/cartesi/rollups-node/blob/8b86538ea54f0dfca0fe8b64c26b90c3250bb02a/pkg/ethutil/ethutil.go#L122) to retrieve the info you want to log.

fmoura: This PR ended up with two different purposes. On would be to add verbose log, to check the steps on adding an Input. the other would be have a better feedback ( more data ) about the input. 

As the former was not fully implemented, I will go with your suggestion and make this Event Logging at the client code

marcelstanley (dismissed):

marcelstanley:
@fmoura, please change the commit message to `feat(cli): add '--verbose' option to send command`

marcelstanley (approved):

fmoura:
Closed in favor of #521

PR #498, opened 2024-07-01
feat(advancer): add tooling for creating snapshots programmatically

renan061:


GMKrieger (dismissed):
LGTM

fmoura:
Maybe current `/internal/machine/snapshot.go` that takes a snapshot from a previous existing machine docker image should be moved to this new tooling package. It has the same functionality, and still can be useful when testing against actual applications

fmoura (dismissed):

test/snapshot/snapshot.go
```diff
@@ -0,0 +1,113 @@
...
+	"strings"
+
+	"github.com/cartesi/rollups-node/pkg/emulator"
+)
+
+const (
+	ramLength = 64 << 20
+)
+
+var ImagesPath = "/usr/share/cartesi-machine/images/"
```
marcelstanley, line 21: This is treated as a constant in the code

test/snapshot/snapshot.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (snapshot *Snapshot) createTempDir() error {
+	temp, err := os.MkdirTemp("", snapshot.pkg+"-*")
+	if err != nil {
+		return err
+	}
+	snapshot.temp = temp
+	snapshot.Dir = snapshot.temp + "/snapshot"
+	return nil
+}
```
marcelstanley, line 86: Perhaps we could simplify this, get rid of `temp` and have only `Dir` , which would be the actual temporary directory where the snapshot will be stored

test/snapshot/snapshot.go
renan061: It is global that can be reset if the images are located in another folder.
I've added documentation to explain that.

test/snapshot/snapshot.go
renan061: We need to have some directory to create the snapshot into, this is `temp`.
For us to merge `dir` and `temp`, we would need to move the files from "snapshot/" into "temp/".
This would add another time-consuming OS operation.

Also, other (future) methods for creating snapshots (e.g. FromFile, FromCode) benefit from this separate folder.

renan061:
> Maybe current `/internal/machine/snapshot.go` that takes a snapshot from a previous existing machine docker image should be moved to this new tooling package. It has the same functionality, and still can be useful when testing against actual applications

Yeah, we should do this. I'm adding it to the technical debt list.

marcelstanley (dismissed):

GMKrieger (approved):

fmoura (approved):

vfusco (approved):

PR #500, opened 2024-07-02
fix: Fix Devnet initialization config

fmoura:
Small fix. It can make some tests fail

### DIFF
--- a/internal/deps/deps.go
+++ b/internal/deps/deps.go
@@ -28,7 +28,7 @@ const (
 	DefaultDevnetDockerImage             = "cartesi/rollups-node-devnet:devel"
 	DefaultDevnetPort                    = "8545"
 	DefaultDevnetBlockTime               = "1"
-	DefaultDevnetBlockToWaitForOnStartup = "21"
+	DefaultDevnetBlockToWaitForOnStartup = "17"
 	DefaultDevnetNoMining                = false
 
 	numPostgresCheckReadyAttempts = 2


pkg/ethutil/ethutil_test.go
```diff
@@ -62,32 +63,52 @@ func (s *EthUtilSuite) TestAddInput() {
...
+	waitGroup := sync.WaitGroup{}
+	waitGroup.Add(1)
+
+	go func() {
+		waitGroup.Done()
+		inputIndex, err := AddInput(s.ctx, s.client, s.book, s.signer, payload)
+		if err != nil {
+			errChan <- err
+			return
+		}
```
GMKrieger, line 78: You should use `!s.Nil(err)` to properly use the suite

GMKrieger (commented):
There are a couple of lint problems making the CI fail, aside from that, LGTM

pkg/ethutil/ethutil_test.go
fmoura: Suite functions do not work inside Go routines. We need to "send them out" to threat them

GMKrieger (dismissed):

marcelstanley (commented):
I'd suggest changing the commit message for 96c04c249c48ef66012ee2284f1e27649fcd1d40 to `chore(devnet): update default initial block`.

BTW, what's the reason for the change? Is it the update to contracts 2.0? Whatever it is, please add it to the commit details.

pkg/ethutil/ethutil_test.go
marcelstanley: The change was removed from the PR, @GMKrieger

GMKrieger (approved):

vfusco (approved):

fmoura:
This PR is outdated.

PR #501, opened 2024-07-02
fix: Fix eth util tests

fmoura:
These tests fails sometimes. Now they are adapted to use "Devnet no mining" feature

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -211,6 +211,8 @@ jobs:
           version: v1.58.2
 
       - name: Run Go tests
+        env:
+          TESTCONTAINERS_RYUK_DISABLED: true
         run: go test ./...
 
   build-docker:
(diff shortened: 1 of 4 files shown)


GMKrieger (changes requested):
There are overlapping commits between this PR and #500. The way it is now, this PR is redundant as the #500 has both commits.

GMKrieger (approved):

.github/workflows/build.yml
```diff
@@ -211,6 +211,8 @@ jobs:
           version: v1.58.2
 
       - name: Run Go tests
+        env:
+          TESTCONTAINERS_RYUK_DISABLED: true
```
marcelstanley, line 215: We could move this to a separate commit

marcelstanley (commented):
Besides adding a separate commit, we could rephrase the current commit as:
- replace `fix(deps): Update testcontainers usage` with
`fix(tests): update port usage for testcontainers`
- replace`fix(util): Fix ethutil tests` with `fix(tests): mine blocks while adding inputs`

If possible, add more context to them as well

marcelstanley (approved):

PR #502, opened 2024-07-02
Feature/input reader integration tests

fmoura:


fmoura:
By Default Anvil takes 64 blocks to mark the first block as finalized. This by default takes at least 1:04 minutes ( one block per second). This is not feasible for Automated Integration Tests.

I was not successful using the Devnet "no mining option"  due to the needed synchronization. 

The other possibility would be use the new Anvil configuration called `slots_in_a_epoch`. But this configuration was added to Foundry April 1st nightly build, and if Anvil version from this one on is used, our code tooling to add inputs fails with error 
```
fail to recover transaction: invalid transaction v, r, s values
```

So a little more time is needed to overcome this obstacles

PR #519, opened 2024-07-15
Add checks to numeric columns in the repository

GMKrieger:
The Node repository uses Numeric to represent uint64 values since Postgresql doesn't have a proper type for unsigned ints.

To keep the data consistent, checks have been added to all numeric columns to verify if the values comply with the uint64 limits.

### DIFF
--- a/internal/repository/migrations/000001_create_application_input_claim_output_report_nodeconfig.up.sql
+++ b/internal/repository/migrations/000001_create_application_input_claim_output_report_nodeconfig.up.sql
@@ -9,13 +9,18 @@ CREATE TYPE "ClaimStatus" AS ENUM ('PENDING', 'SUBMITTED', 'FINALIZED');
 
 CREATE TYPE "DefaultBlock" AS ENUM ('FINALIZED', 'LATEST', 'PENDING', 'SAFE');
 
+CREATE FUNCTION public.f_maxuint64()
+  RETURNS NUMERIC(20,0)
+  LANGUAGE sql IMMUTABLE PARALLEL SAFE AS
+'SELECT 18446744073709551615';
+
 CREATE TABLE "application"
 (
     "id" SERIAL,
     "contract_address" BYTEA NOT NULL,
     "template_hash" BYTEA NOT NULL,
     "snapshot_uri" VARCHAR(4096) NOT NULL,
-    "last_processed_block" NUMERIC(20,0) NOT NULL,
+    "last_processed_block" NUMERIC(20,0) NOT NULL CHECK ("last_processed_block" >= 0 AND "last_processed_block" <= f_maxuint64()),
     "status" "ApplicationStatus" NOT NULL,
     "epoch_length" INT NOT NULL,
     CONSTRAINT "application_pkey" PRIMARY KEY ("id"),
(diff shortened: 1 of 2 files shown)


internal/repository/migrations/000001_create_application_input_claim_output_report_nodeconfig.up.sql
```diff
@@ -15,7 +15,7 @@ CREATE TABLE "application"
     "contract_address" BYTEA NOT NULL,
     "template_hash" BYTEA NOT NULL,
     "snapshot_uri" VARCHAR(4096) NOT NULL,
-    "last_processed_block" NUMERIC(20,0) NOT NULL,
+    "last_processed_block" NUMERIC(20,0) NOT NULL CHECK ("last_processed_block" >= 0 AND "last_processed_block" <= 18446744073709551615),
```
marcelstanley, line 18: I'm wondering if we could create a constant for `18446744073709551615` and reuses throughout the code.

GMKrieger: I checked if this was possible before doing it this way, but postgres has only three types of constants, none of which would support this large number https://www.postgresql.org/docs/7.0/syntax543.htm

To implement a constant like solution, we would need to use a function, which makes things more complicated, imho https://stackoverflow.com/questions/13316773/is-there-a-way-to-define-a-named-constant-in-a-postgresql-query

fmoura: I tried this 
```
CREATE FUNCTION public.f_maxuint64()
  RETURNS NUMERIC(20,0)
  LANGUAGE sql IMMUTABLE PARALLEL SAFE AS
'SELECT 18446744073709551615';


CREATE TABLE "application"
(
    "id" SERIAL,
    "contract_address" BYTEA NOT NULL,
    "template_hash" BYTEA NOT NULL,
    "snapshot_uri" VARCHAR(4096) NOT NULL,
    "last_processed_block" NUMERIC(20,0) NOT NULL CHECK ("last_processed_block" >= 0 AND "last_processed_block" <= f_maxuint64() ),
    "status" "ApplicationStatus" NOT NULL,
    "epoch_length" INT NOT NULL,
    CONSTRAINT "application_pkey" PRIMARY KEY ("id"),
    UNIQUE("contract_address")
);
```
And seems to work

internal/repository/migrations/000001_create_application_input_claim_output_report_nodeconfig.up.sql
```diff
@@ -9,13 +9,18 @@ CREATE TYPE "ClaimStatus" AS ENUM ('PENDING', 'SUBMITTED', 'FINALIZED');
...
 CREATE TYPE "DefaultBlock" AS ENUM ('FINALIZED', 'LATEST', 'PENDING', 'SAFE');
 
+CREATE FUNCTION public.f_maxuint64()
+  RETURNS NUMERIC(20,0)
+  LANGUAGE sql IMMUTABLE PARALLEL SAFE AS
+'SELECT 18446744073709551615';
+
 CREATE TABLE "application"
 (
     "id" SERIAL,
```
fmoura, line 19: Maintaining it as "Serial" and not change to "Bigserial" is intended, right? As the number of applications at the same node should not be that big

internal/repository/migrations/000001_create_application_input_claim_output_report_nodeconfig.up.sql
marcelstanley: I was wondering if a function like this one would be optimized by postgres and what would be its effect in terms of performance. I guess [this explanation from the postgres docs](https://www.postgresql.org/docs/current/xfunc-optimization.html)is good enough, though.

marcelstanley (approved):

internal/repository/migrations/000001_create_application_input_claim_output_report_nodeconfig.up.sql
GMKrieger: Yes, it's just as you said. No need to use `bigserial` here, unless we expect more than 2147483647 applications per node,

fmoura (approved):

PR #532, opened 2024-07-22
Port mine command to 1.x

marcelstanley:
- Cherry-picks command and devnet changes made at c7193114c7ef6fe254453d96b735a8b4de351a89 for next/2.0.
- Also extends the command to allow for mining multiple blocks at a time. 

As the cli is solely used for _devnet_, it assumes the provider is foundry/anvil in order to perform that instead of relying on method `rpc_modules` to discover if `amvil_mine` can be used instead of a loop of `evm_mine` calls to achieve the same result.

### DIFF
--- a/cmd/cartesi-rollups-cli/root/deps/deps.go
+++ b/cmd/cartesi-rollups-cli/root/deps/deps.go
@@ -51,8 +51,12 @@ func init() {
 		"Devnet local listening port number")
 
 	Cmd.Flags().StringVar(&depsConfig.Devnet.BlockTime, "devnet-block-time",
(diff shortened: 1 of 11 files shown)


vfusco (approved):

GMKrieger (approved):
LGTM, just one question for a possible future issue: since this is the new way to advance the blockchain state, does it make sense to still have the `increasetime` command? Should we delete it?

marcelstanley:
> LGTM, just one question for a possible future issue: since this is the new way to advance the blockchain state, does it make sense to still have the `increasetime` command? Should we delete it?

You're right, this should be revisited later on.

PR #533, opened 2024-07-25
feat(deps): add devnet-slots-in-an-epoch argument

fmoura:


cmd/cartesi-rollups-cli/root/deps/deps.go
```diff
@@ -56,6 +56,10 @@ func init() {
 		deps.DefaultDevnetBlockTime,
 		"Devnet mining block time")
 
+	Cmd.Flags().StringVar(&depsConfig.Devnet.SlotsInAnEpoch, "devnet-slots-in-an-epoch",
+		deps.DefaultSlotsInAnEpoch,
+		"Devnet slots in an epoch")
```
marcelstanley, line 61: What's the exact purpose of adding this new argument?

Also, perhaps adding a longer explanation would make it easier to understand that we're referring to. See explanation [available in anvil](https://book.getfoundry.sh/reference/anvil/):
> To speed up the finalization of blocks, you can use the --slots-in-an-epoch flag with a value of 1 for example. This will lead to the block at height N-2 being finalized, where N is the latest block.

BTW, how does this work with `CARTESI_BLOCKCHAIN_FINALITY_OFFSET`?

fmoura: I will improve the docs. 

The purpose is to , when testing Node against Devnet, and setting the Default Block to "finalized" you do not have to wait for the Anvil's default  64 blocks long epoch ( which will take about a minute to finalize)

fmoura: About the `CARTESI_BLOCKCHAIN_FINALITY_OFFSET`. It is a "legacy" parameter, still in use by "Claimer". It has the same meaning as Anvil's `slots-in-an-epoch`, but not sure what are it's effects on Claimer behavior ( probably how many blocks it waits to consider claim submission transaction as finalized)

vfusco:
Deprecated

PR #535, opened 2024-07-26
Check database sslmode when validating schema

### LINKED ISSUE #412: Add sslmode configuration to startup
## 📚 Context

During development, a conflict occurred on how to correctly deploy the migrations. The local database we run is not enabled with ssl, and does not require certificates, different from the production database. The migrations driver requires a command to inform if the database is configured with ssl or not, and will not perform if this configuration is wrong.

The problem is even deployed to production with the correct configuration, the tests will break, as the database used for the tests does not used ssl, as well. There needs to be a way to differentiate when to use the sslmode.

## ✔️ Solution

The Node needs to know to where it's being deployed, either locally or production. There are 2 ways to this: either the Node is said explicitly, and there needs to be flag to tell it, or it needs to check certain configurations that are known to be different and infer the deployment.

The second solution is leaner and avoids bogging the startup command, but runs the risk of breaking in the future if the configuration changes and the check is not performed correctly.

From the moment of writing this issue, there isn't an obvious path to follow. Part of the work needed is studying the Node and convening with the team to chose a way to move forward.

## 📈 Subtasks

- [ ] Find a method to differentiate local and production startups
- [ ] Add the different ssl mode configuration to each startup

GMKrieger:
Closes #412

### DIFF
--- a/internal/node/startup/startup.go
+++ b/internal/node/startup/startup.go
@@ -19,10 +19,23 @@ import (
 )
 
 // Validates the Node Database Schema Version
-func ValidateSchema(postgresEndpoint string) error {
-	schemaManager, err := repository.NewSchemaManager(postgresEndpoint)
-	if err != nil {
-		return err
+func ValidateSchema(config config.NodeConfig) error {
+	var (
+		schemaManager *repository.SchemaManager
+		err           error
+	)
+
+	if !config.PostgresSslMode {
+		schemaManager, err = repository.NewSchemaManager(
+			fmt.Sprintf("%v?sslmode=disable", config.PostgresEndpoint.Value))
+		if err != nil {
+			return err
+		}
+	} else {
+		schemaManager, err = repository.NewSchemaManager(config.PostgresEndpoint.Value)
+		if err != nil {
+			return err
+		}
 	}
 	defer schemaManager.Close()
 	err = schemaManager.ValidateSchemaVersion()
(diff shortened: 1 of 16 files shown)


marcelstanley:
I'd suggest changing [chore: remove obsolete config from yaml files](https://github.com/cartesi/rollups-node/pull/535/commits/a0958f549fc856d9d54c8133e47f51f43f7c155e) to `chore(build): remove obsolete compose config`

internal/node/startup/startup.go
```diff
@@ -19,10 +20,23 @@ import (
...
-	schemaManager, err := repository.NewSchemaManager(postgresEndpoint)
-	if err != nil {
-		return err
+func ValidateSchema(config config.NodeConfig) error {
+	var (
+		schemaManager *repository.SchemaManager
+		err           error
+	)
+
+	if strings.Contains(config.BlockchainHttpEndpoint.Value, "devnet") {
```
marcelstanley, line 29: I'm not sure that checking for [`devnet`](https://github.com/cartesi/rollups-node/blob/584edb0145ae17efbea577f2114684c54dddf773/build/compose-devnet.yaml#L18) is robust enough. Perhaps we should add an extra config parameter to explicitly ask to disable SSL.

vfusco: I agree

vfusco (changes requested):
I agree with @marcelstanley 's comment

fmoura (dismissed):

marcelstanley (changes requested):

fmoura (changes requested):

fmoura (commented):
LGTM 

But there is an error on Github Actions

fmoura (approved):

marcelstanley (approved):

vfusco (approved):

PR #548, opened 2024-08-01
feat(database): Add Epoch and Snapshot tables

### LINKED ISSUE #544: [DB] Add Epoch and Snapshot tables
## 📚 Context

Node needs to ensure proper claim handling and application execution between restarts.

## ✔️ Solution

Update Database schema to add Epoch and Snapshot tables

fmoura:
closes #544

### DIFF
--- a/internal/evmreader/evmreader.go
+++ b/internal/evmreader/evmreader.go
@@ -9,7 +9,7 @@ import (
 	"log/slog"
 	"math/big"
 
-	"github.com/cartesi/rollups-node/internal/node/model"
+	. "github.com/cartesi/rollups-node/internal/node/model"
 	"github.com/cartesi/rollups-node/pkg/contracts/inputbox"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
@@ -18,52 +18,38 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
-type (
-	Address              = common.Address
-	Application          = model.Application
-	Input                = model.Input
-	DefaultBlock         = model.DefaultBlock
-	NodePersistentConfig = model.NodePersistentConfig
-	InputBoxInputAdded   = inputbox.InputBoxInputAdded
-	FilterOpts           = bind.FilterOpts
-	Context              = context.Context
-	Header               = types.Header
-	Subscription         = ethereum.Subscription
-	Epoch                = model.Epoch
-)
-
 // Interface for Input reading
 type InputSource interface {
 	// Wrapper for FilterInputAdded(), which is automatically generated
 	// by go-ethereum and cannot be used for testing
 	RetrieveInputs(
-		opts *FilterOpts,
-		appContract []Address,
+		opts *bind.FilterOpts,
+		appContract []common.Address,
 		index []*big.Int,
-	) ([]InputBoxInputAdded, error)
+	) ([]inputbox.InputBoxInputAdded, error)
 }
 
 // Interface for the node repository
 type EvmReaderRepository interface {
 	InsertInputsAndUpdateLastProcessedBlock(
-		ctx Context,
+		ctx context.Context,
 		inputs []Input,
 		blockNumber uint64,
-		appAddress Address,
+		appAddress common.Address,
 	) error
 	GetAllRunningApplications(
-		ctx Context,
+		ctx context.Context,
 	) ([]Application, error)
 	GetNodeConfig(
-		ctx Context,
+		ctx context.Context,
 	) (*NodePersistentConfig, error)
 	GetEpoch(
-		ctx Context,
+		ctx context.Context,
 		indexKey uint64,
-		appAddressKey Address,
+		appAddressKey common.Address,
 	) (*Epoch, error)
 	InsertEpoch(
-		ctx Context,
+		ctx context.Context,
 		epoch *Epoch,
 	) (uint64, error)
 }
@@ -72,18 +58,18 @@ type EvmReaderRepository interface {
 // interface needed by the EvmReader. It must be bound to an HTTP endpoint
 type EthClient interface {
 	HeaderByNumber(
-		ctx Context,
+		ctx context.Context,
 		number *big.Int,
-	) (*Header, error)
+	) (*types.Header, error)
 }
 
 // EthWsClient mimics part of ethclient.Client functions to narrow down the
 // interface needed by the EvmReader. It must be bound to a WS endpoint
 type EthWsClient interface {
 	SubscribeNewHead(
-		ctx Context,
-		ch chan<- *Header,
-	) (Subscription, error)
+		ctx context.Context,
+		ch chan<- *types.Header,
+	) (ethereum.Subscription, error)
 }
 
 type SubscriptionError struct {
@@ -144,10 +130,10 @@ func (r *EvmReader) Run(
 // Watch for new blocks and reads new inputs based on the
 // default block configuration, which have not been processed yet.
 func (r *EvmReader) watchForNewBlocks(
-	ctx Context,
+	ctx context.Context,
 	ready chan<- struct{},
 ) error {
-	headers := make(chan *Header)
+	headers := make(chan *types.Header)
 	sub, err := r.wsClient.SubscribeNewHead(ctx, headers)
 	if err != nil {
 		return fmt.Errorf("could not start subscription: %v", err)
@@ -177,7 +163,7 @@ func (r *EvmReader) watchForNewBlocks(
 }
 
 // Check if is there new Inputs for all running Applications
-func (r *EvmReader) checkForNewInputs(ctx Context) error {
+func (r *EvmReader) checkForNewInputs(ctx context.Context) error {
 
 	// Get All Applications
 	apps, err := r.repository.GetAllRunningApplications(ctx)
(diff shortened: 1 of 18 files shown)


internal/evmreader/evmreader.go
```diff
@@ -29,6 +30,7 @@ type (
 	Context              = context.Context
 	Header               = types.Header
 	Subscription         = ethereum.Subscription
+	Epoch                = model.Epoch
 )
```
renan061, line 34: This idea of defining aliases for types to shorten their names is much "not-Go-like".
I would prefer if we kept the Go standard of having types prefixed by their packages.

internal/evmreader/evmreader.go
```diff
@@ -287,9 +298,34 @@ func (r *EvmReader) readInputs(
...
 		filter = append(filter, app.ContractAddress)
 		inputsMap[app.ContractAddress] = []Input{}
+
+		// TEMPORARY Get the Default Epoch until epochs are not properly handled
+		epoch, err := r.repository.GetEpoch(ctx, 0, app.ContractAddress)
+		if err != nil {
+			return err
+		}
+		if epoch == nil {
+			// Create a Default Epoch
```
renan061, line 312: I don't understand this.

internal/node/model/models.go
```diff
@@ -46,39 +41,60 @@ const (
 	DefaultBlockStatusSafe      DefaultBlock = "SAFE"
 )
 
+const (
```
renan061, line 44: What does `EpochStatusReceivedLastInput` mean? Is that the same as "epoch closed"? Why not call it that?

Also, I think `EpochStatusClaimComputed` is better than `EpochStatusCalculatedClaim`.

And we could do the same reversion for the other names: "ClaimSubmitted", "ClaimAccepted", ...

internal/node/model/models.go
```diff
@@ -11,10 +11,11 @@ import (
 type (
```
renan061, line 11: This is an exception, because we are reexporting these names.
Leaving these alias as is is fine.

internal/repository/base.go
```diff
@@ -121,6 +118,53 @@ func (pg *Database) InsertApplication(
 	return nil
 }
 
+func (pg *Database) InsertEpoch(
+	ctx context.Context,
+	epoch *Epoch,
+) (uint64, error) {
+
+	var (
```
renan061, line 126: One line: `var id uint64`

internal/repository/base.go
```diff
@@ -545,62 +651,58 @@ func (pg *Database) GetReport(
...
 	FROM
-		claim
+		snapshot s
+	INNER JOIN
+		input i
+	ON
+		i.id = s.input_id
 	WHERE
-		application_address=@appAddress and index=@index`
+		s.application_address=@appAddress and i.index=@inputIndex
```
renan061, line 678: `AND` uppercase.

internal/repository/evmreader.go
```diff
@@ -57,21 +59,22 @@ func (pg *Database) InsertInputsAndUpdateLastProcessedBlock(
 			"rawData":     input.RawData,
 			"blockNumber": input.BlockNumber,
 			"appAddress":  input.AppAddress,
+			"epochId":     input.EpochId,
 		}
 		_, err = tx.Exec(ctx, query, inputArgs)
 		if err != nil {
-			return fmt.Errorf("%w: %w", errInsertInputs, err)
+			return errors.Join(fmt.Errorf("%w: %w", errInsertInputs, err), tx.Rollback(ctx))
```
renan061, line 66: You can remove the `fmt.Errorf` call by doing `errors.Join(errInsertInputs, err, tx.Rollback(ctx))`.

internal/repository/migrations/000001_create_application_input_claim_output_report_nodeconfig.up.sql
```diff
@@ -5,10 +5,10 @@ CREATE TYPE "ApplicationStatus" AS ENUM ('RUNNING', 'NOT RUNNING');
 
 CREATE TYPE "InputCompletionStatus" AS ENUM ('NONE', 'ACCEPTED', 'REJECTED', 'EXCEPTION', 'MACHINE_HALTED', 'CYCLE_LIMIT_EXCEEDED', 'TIME_LIMIT_EXCEEDED', 'PAYLOAD_LENGTH_LIMIT_EXCEEDED');
 
-CREATE TYPE "ClaimStatus" AS ENUM ('PENDING', 'SUBMITTED', 'FINALIZED');
-
 CREATE TYPE "DefaultBlock" AS ENUM ('FINALIZED', 'LATEST', 'PENDING', 'SAFE');
 
+CREATE TYPE "EpochStatus" AS ENUM ('RECEIVING_INPUTS', 'RECEIVED_LAST_INPUT', 'PROCESSED_ALL_INPUTS', 'CALCULATED_CLAIM', 'SUBMITTED_CLAIM', 'ACCEPTED_CLAIM', 'REJECTED_CLAIM');
```
renan061, line 10: See `model` comment.

internal/repository/validator.go
```diff
@@ -130,60 +131,89 @@ func (pg *Database) getAllOutputsFromProcessedInputs(
 	return nil, nil
 }
 
-func (pg *Database) FinishEpoch(
+func (pg *Database) SetEpochClaimAndInsertProofsTransaction(
```
renan061, line 134: Is this going to be changed again in @torives' PR?

renan061 (changes requested):
First review (haven't reviewed the tests.)

internal/node/model/models.go
```diff
@@ -46,39 +41,60 @@ const (
...
+	Id              uint64
+	AppAddress      Address
+	Index           uint64
+	FirstBlock      uint64
+	LastBlock       uint64
+	ClaimHash       *Hash
+	TransactionHash *Hash
+	Status          EpochStatus
+}
+
```
GMKrieger, line 81: For formatting sake, let's put the foreign keys as the last values on the struct

internal/node/model/models.go
```diff
@@ -87,15 +103,13 @@ type Output struct {
...
-	Index                uint64
-	Status               ClaimStatus
-	OutputMerkleRootHash Hash
-	TransactionHash      *Hash
-	AppAddress           Address
+type Snapshot struct {
+	Id         uint64
+	InputId    uint64
+	AppAddress Address
+	URI        string
```
GMKrieger, line 114: Same thing here

internal/repository/base.go
```diff
@@ -220,30 +267,24 @@ func (pg *Database) InsertReport(
 	return nil
 }
 
-func (pg *Database) InsertClaim(
+func (pg *Database) InsertSnapshot(
 	ctx context.Context,
-	claim *Claim,
+	snapshot *Snapshot,
```
GMKrieger, line 272: This is a good opportunity to add `RETURNING` to this method, since it's basically brand new

internal/repository/migrations/000001_create_application_input_claim_output_report_nodeconfig.up.sql
```diff
@@ -83,11 +85,25 @@ CREATE TABLE "report"
...
     "default_block" "DefaultBlock" NOT NULL,
     "input_box_deployment_block" INT NOT NULL,
     "input_box_address" BYTEA NOT NULL,
     "chain_id" INT NOT NULL,
-    "iconsensus_address" BYTEA NOT NULL
+    "iconsensus_address" BYTEA NOT NULL,
+    "epoch_length" INT NOT NULL
 );
+
+
```
GMKrieger, line 109: One too many break lines

internal/repository/migrations/000002_create_postgraphile_view.up.sql
```diff
@@ -48,30 +61,15 @@ CREATE OR REPLACE VIEW graphql."reports" AS
...
   E'@foreignKey (input_index) references inputs(index)|@fieldName inputByInputIndex';
 
 COMMENT ON VIEW graphql."reports" is
   E'@foreignKey (input_index) references inputs(index)|@fieldName inputByInputIndex';
 
-COMMENT ON VIEW graphql."claims" is
-  E'@foreignKey (output_index) references outputs(index)|@fieldName outputByOutputIndex\n@foreignKey (application_address) references applications(contract_address)|@fieldName applicationByApplicationAddress';
\ No newline at end of file
+COMMENT ON VIEW graphql."epochs" is
+  E'@foreignKey (application_address) references applications(contract_address)|@fieldName applicationByApplicationAddress';
```
GMKrieger, line 75: Missing EOF break line

internal/repository/validator.go
GMKrieger, line 1: As @renan061 pointed out, this will be changed very soon. I will refrain from reviewing for now.

GMKrieger (changes requested):
I was going to comment that the snapshot table doesn't test fail cases, but reading the models, it doesn't have one.

We need to make some changes. We accept same app address and input? If yes, we need to change how we get snapshots. If no, we need to add an unique constraint to this columns.

internal/repository/validator.go
fmoura: This is already aligned with @torives . This is the actual name.

internal/evmreader/evmreader.go
fmoura: Removed

internal/repository/validator.go
torives: > Is this going to be changed again in @torives' PR?

In an upcoming PR tackling #549, right @fmoura?

internal/evmreader/evmreader.go
torives: > I would prefer if we kept the Go standard of having types prefixed by their packages. 

The only exception we currently have is the `model` package. We're using ["dot imports"](https://go.dev/ref/spec#Import_declarations) to add its exported names to the local namespace and avoid prefixing everything with `model.`. [Here](https://github.com/cartesi/rollups-node/blob/next/2.0/internal/repository/validator.go#L11), for example.

fmoura: Done!

internal/node/model/models.go
fmoura: Changed

internal/repository/base.go
fmoura: Done

internal/repository/base.go
fmoura: Done

internal/repository/evmreader.go
fmoura: Done

internal/repository/migrations/000001_create_application_input_claim_output_report_nodeconfig.up.sql
fmoura: Changed

fmoura:
> I was going to comment that the snapshot table doesn't test fail cases, but reading the models, it doesn't have one.
> 
> We need to make some changes. We accept same app address and input? If yes, we need to change how we get snapshots. If no, we need to add an unique constraint to this columns.

Changes the `report`, `outuput`and `snapshot` constraints to enforce unique `index` per applications

GMKrieger (approved):

vfusco (approved):

renan061 (approved):

Issue #549, opened 2024-08-02
[DB] Add Validator repository methods for epoch handling


fmoura:
## 📚 Context

New Epoch handling rose some new needs to the Validator Component

## ✔️ Solution

Implement new Repository methods needed by Validator components

PR #560, opened 2024-08-09
feat(repository): update validator methods

torives:
I've returned the ID for newly inserted inputs and outputs to make my life easier, before realizing we're not enforcing their index uniqueness by app yet. I left it here since we will do it for all insert functions sometime soon.

Closes https://github.com/cartesi/rollups-node/issues/549.

### DIFF
--- a/internal/evmreader/evmreader.go
+++ b/internal/evmreader/evmreader.go
@@ -36,7 +36,7 @@ type EvmReaderRepository interface {
 		inputs []Input,
 		blockNumber uint64,
 		appAddress common.Address,
-	) error
+	) ([]uint64, error)
 	GetAllRunningApplications(
 		ctx context.Context,
 	) ([]Application, error)
@@ -326,7 +326,7 @@ func (r *EvmReader) readInputs(
 				len(inputs),
 			)
 		}
-		err = r.repository.InsertInputsAndUpdateLastProcessedBlock(
+		_, err = r.repository.InsertInputsAndUpdateLastProcessedBlock(
 			ctx,
 			inputs,
 			endBlock,
--- a/internal/repository/evmreader_test.go
+++ b/internal/repository/evmreader_test.go
@@ -10,7 +10,6 @@ import (
 
 func (s *RepositorySuite) TestInsertInputsAndUpdateLastProcessedBlock() {
 	input0 := Input{
-		Id:               6,
 		Index:            5,
 		CompletionStatus: InputStatusNone,
 		RawData:          common.Hex2Bytes("deadbeef"),
@@ -20,7 +19,6 @@ func (s *RepositorySuite) TestInsertInputsAndUpdateLastProcessedBlock() {
 	}
 
 	input1 := Input{
-		Id:               7,
 		Index:            6,
 		CompletionStatus: InputStatusNone,
 		RawData:          common.Hex2Bytes("deadbeef"),
@@ -29,17 +27,17 @@ func (s *RepositorySuite) TestInsertInputsAndUpdateLastProcessedBlock() {
 		EpochId:          1,
 	}
 
-	var inputs []Input
-	inputs = append(inputs, input0)
-	inputs = append(inputs, input1)
-
-	err := s.database.InsertInputsAndUpdateLastProcessedBlock(
+	ids, err := s.database.InsertInputsAndUpdateLastProcessedBlock(
 		s.ctx,
-		inputs,
+		[]Input{input0, input1},
 		6,
 		common.HexToAddress("deadbeef"),
 	)
 	s.Require().Nil(err)
+	s.Require().Len(ids, 2)
+
+	input0.Id = ids[0]
+	input1.Id = ids[1]
 
 	response, err := s.database.GetInput(s.ctx, 5, common.HexToAddress("deadbeef"))
 	s.Require().Nil(err)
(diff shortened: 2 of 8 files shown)


internal/repository/validator.go
```diff
@@ -7,198 +7,339 @@ import (
...
 )
 
-const DefaultServiceTimeout = 5 * time.Minute
+var NilEpoch = errors.New("epoch cannot be nil")
 
-func (pg *Database) GetLastProcessedBlock(
+// GetOutputs returns outputs produced by inputs sent to the application
+// between start and end blocks, inclusive. Outputs are returned in ascending
+// order by index.
+func (pg *Database) GetOutputs(
```
GMKrieger, line 21: I think the name `GetOutputs` is too generic, it should be `GetOutputsBetweenBlocks` or something similar

internal/repository/evmreader_test.go
```diff
@@ -43,6 +41,8 @@ func (s *RepositorySuite) TestInsertInputsAndUpdateLastProcessedBlock() {
 
 	response, err := s.database.GetInput(s.ctx, 5, common.HexToAddress("deadbeef"))
 	s.Require().Nil(err)
+	// necessary while the insert function does not return the generated Id
+	response.Id = 0
```
GMKrieger, line 45: Since you already updated the base insert functions, I think it's ok to update the get function as well to return the id

internal/repository/validator.go
```diff
@@ -7,198 +7,339 @@ import (
...
+	query := `
+	SELECT
+		id,
+    	application_address,
+        index,
+        first_block,
+		last_block,
+		claim_hash,
+		transaction_hash,
+		status
```
GMKrieger, line 93: Some of the lines have wrong indentation

internal/repository/validator.go
```diff
@@ -7,198 +7,339 @@ import (
...
+		input
+	WHERE
+		epoch_id = %d
+	ORDER BY
+		index DESC
+	LIMIT 1
+	`, epoch.Id)
+	var outputHash Hash
+
+	err = pg.db.QueryRow(ctx, query).Scan(&outputHash)
```
GMKrieger, line 182: You should use prepared statements and named args here to keep consistent with the rest of the code

internal/repository/validator.go
```diff
@@ -7,198 +7,339 @@ import (
...
-		o.input_id,
-		i.status
+		id,
+    	application_address,
+        index,
+        first_block,
+		last_block,
+		claim_hash,
+		transaction_hash,
+		status
```
GMKrieger, line 217: Wrong indentation here as well

internal/repository/validator.go
```diff
@@ -7,198 +7,339 @@ import (
...
+//
+// 1. Updates an epoch, adding its claim and modifying its status.
+//
+// 2. Updates several outputs with their Keccak256 hash and proof.
 func (pg *Database) SetEpochClaimAndInsertProofsTransaction(
 	ctx context.Context,
-	epoch Epoch,
-	outputs []Output,
+	epoch *Epoch,
+	outputs []*Output,
```
GMKrieger, line 274: Do we need an array of pointers here? If there's a nil value in here, shouldn't we just remove it from the array?

GMKrieger (changes requested):

internal/repository/validator.go
torives: There's a discussion to be had about standardizing all these getter methods, but let's leave it for later. 👌🏽 I'll use `GetOutputsProducedInBlockRange`. Done in 16ea3f0a22d2cf364f83694b975ca45e00f68e24.

internal/repository/evmreader_test.go
torives: Done in 24bbce4dd953a53a0e6a018952ce3bba50268b6d.

internal/repository/validator.go
torives: Something funny is happening with Github's rendering when spaces are mixed with tabs, it looks as it should in my editor. 🤷🏽‍♂️ . But I'll standardize it anyway 😉. Done in fcb5cc35fdf99dac9b5a1dd270cf21926c1083e5.

internal/repository/validator.go
torives: Done in 9d1b31813a6d74c82792ff5cf8a170cbb0912d14

internal/repository/validator.go
torives: Done in 59b274d212eeba3100d893bd0df2cd9a02987282

internal/repository/validator.go
fmoura: Slices copies share the underline array, right? So copying slices should not be expensive.

Using actual instances instead of pointers will prevent these possible caller errors

internal/evmreader/evmreader.go
```diff
@@ -36,7 +36,7 @@ type EvmReaderRepository interface {
 		inputs []Input,
 		blockNumber uint64,
 		appAddress common.Address,
-	) error
+	) ([]uint64, error)
```
fmoura, line 39: Evm Reader Repository Mock needs to be updated to conform to new the new method signature

fmoura (changes requested):

GMKrieger (approved):

fmoura (approved):

PR #580, opened 2024-08-28
Add inspect package

### LINKED ISSUE #575: Add inspect server package (alpha)
## 📚 Context
The inspect server is the gateway through which we can get reports from the machine. 

## ✔️ Solution
For the 2.0 alpha, we will implement a basic version of it that allows the user to make a simple call.

GMKrieger:
Closes #575

### DIFF
--- a/internal/inspect/inspect.go
+++ b/internal/inspect/inspect.go
@@ -1,0 +1,162 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+package inspect
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+
+	"github.com/cartesi/rollups-node/internal/node/advancer/machines"
+	. "github.com/cartesi/rollups-node/internal/node/model"
+	"github.com/cartesi/rollups-node/internal/nodemachine"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+var (
+	ErrInvalidMachines = errors.New("machines must not be nil")
+	ErrNoApp           = errors.New("no machine for application")
+)
+
+type Inspector struct {
+	machines Machines
+}
+
+type InspectResponse struct {
+	Status     string   `json:"status"`
+	Exception  string   `json:"exception"`
+	Reports    []string `json:"reports"`
+	InputIndex uint64   `json:"processed_input_count"`
+}
+
+// New instantiates a new Inspector.
+func New(machines Machines) (*Inspector, error) {
+	if machines == nil {
+		return nil, ErrInvalidMachines
+	}
+
+	return &Inspector{machines: machines}, nil
+}
+
+func (inspect *Inspector) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var (
+		dapp         Address
+		payload      []byte
+		err          error
+		reports      []string
+		status       string
+		errorMessage string
+	)
+
+	if r.PathValue("dapp") == "" {
+		slog.Info("Bad request",
+			"service", "inspect",
+			"err", "Missing application address")
+		http.Error(w, "Missing application address", http.StatusBadRequest)
+		return
+	}
+
+	dapp = common.HexToAddress(r.PathValue("dapp"))
+	if r.Method == "POST" {
+		payload, err = io.ReadAll(r.Body)
+		if err != nil {
+			slog.Info("Bad request",
+				"service", "inspect",
+				"err", err)
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	} else {
+		if r.PathValue("payload") == "" {
+			slog.Info("Bad request",
+				"service", "inspect",
+				"err", "Missing payload")
+			http.Error(w, "Missing payload", http.StatusBadRequest)
+			return
+		}
+		payload, err = hexutil.Decode(r.PathValue("payload"))
+		if err != nil {
+			slog.Info("Internal server error",
+				"service", "inspect",
+				"err", err)
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	}
+
+	result, err := inspect.process(r.Context(), dapp, payload)
+	if err != nil {
+		slog.Info("Internal server error",
+			"service", "inspect",
+			"err", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	for _, report := range result.Reports {
+		reports = append(reports, hexutil.Encode(report))
+	}
+
+	if result.Accepted {
+		status = "Accepted"
+	} else {
+		status = "Rejected"
+	}
+
+	if result.Error != nil {
+		status = "Exception"
+		errorMessage = fmt.Sprintf("Error on the machine while inspecting: %s", result.Error)
+	}
+
+	response := InspectResponse{
+		Status:     status,
+		Exception:  errorMessage,
+		Reports:    reports,
+		InputIndex: *result.InputIndex,
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	err = json.NewEncoder(w).Encode(response)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/inspect/inspect.go
```diff
@@ -0,0 +1,159 @@
...
+
+var (
+	ErrInvalidMachines   = errors.New("machines must not be nil")
+	ErrInvalidRepository = errors.New("repository must not be nil")
+
+	ErrNoApp      = errors.New("no machine for application")
+	ErrBadRequest = errors.New("inspect bad request")
+)
+
+type Inspect struct {
```
renan061: Keeping with the `advancer`, `reader`, and `validator` standard, we should name this `Inspector`.
Same for the package, the file, etc.

internal/inspect/inspect.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/cartesi/rollups-node/internal/node/advancer/machines"
+	. "github.com/cartesi/rollups-node/internal/node/model"
+	"github.com/cartesi/rollups-node/internal/nodemachine"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+var (
+	ErrInvalidMachines   = errors.New("machines must not be nil")
+	ErrInvalidRepository = errors.New("repository must not be nil")
```
renan061: This is not being used.

internal/inspect/inspect.go
```diff
@@ -0,0 +1,159 @@
...
+func (inspect *Inspect) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var (
+		dapp    Address
+		payload []byte
+		err     error
+		reports []string
+		status  string
+	)
+
+	if r.PathValue("dapp") == "" {
```
renan061, line 57: The request is a POST, we should not use PathValue.
Instead, we should get everything we need from the body of the request, which could be json encoded as in:
`{ "application_address": "...", "query": "..." }`

internal/inspect/inspect.go
```diff
@@ -0,0 +1,159 @@
...
+		payload, err = io.ReadAll(r.Body)
+		if err != nil {
+			slog.Info("Bad request",
+				"service", "inspect",
+				"err", err)
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	} else {
+		if r.PathValue("payload") == "" {
```
renan061, line 76: As before, we should not use PathValue.
If method is not POST, then the endpoint does not exist.

internal/inspect/inspect.go
```diff
@@ -0,0 +1,159 @@
...
+
+// process sends an inspect request to the machine
+func (inspect *Inspect) process(
+	ctx context.Context,
+	app Address,
+	query []byte) (*nodemachine.InspectResult, error) {
+	// Asserts that the app has an associated machine.
+	machine := inspect.machines.GetInspectMachine(app)
+	if machine == nil {
+		panic(fmt.Errorf("%w %s", ErrNoApp, app.String()))
```
renan061: We should return this as an error instead of panicking.

renan061 (dismissed):

internal/inspect/inspect.go
GMKrieger: This method resolves both get and post requests. If it isn't post, it gets the value from the path.

My answer to your other question explains the details better.

internal/inspect/inspect.go
GMKrieger: I agree conceptually, however, this design came to be because I'm trying to emulate the previous API as close as possible, as requested during meetings.

The main problem is with multi dapp, I need to receive both the application address and the payload.

The previous API received an octet-stream on the post requests, so it doesn't make sense to receive any information that's not just the payload bytes.

Therefore, I arrived at this: receive the dapp address on path always, and payload on the path in get requests and a binary on the body in post requests.

internal/inspect/inspect.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type InspectorResponse struct {
+	Status     string   `json:"status"`
+	Exception  string   `json:"exception"`
+	Reports    []string `json:"reports"`
+	InputIndex uint64   `json:"processed_input_count"`
+}
+
+// New instantiates a new Inspect.
```
marcelstanley:
```suggestion
// New instantiates a new Inspector.
```

internal/inspect/inspect.go
marcelstanley: @GMKrieger, I agree with @renan061.

internal/node/services.go
```diff
@@ -99,7 +102,23 @@ func newSupervisorService(
...
+		slog.Error("failed to load the machines", "error", err)
+		os.Exit(1)
+	}
+	defer machines.Close()
+
+	inspect, err := inspect.New(machines)
+	if err != nil {
+		slog.Error("failed to create the inspect", "error", err)
+		os.Exit(1)
+	}
```
marcelstanley:
```suggestion
	// initialize machines for inspect
	repo := &repository.MachineRepository{Database: database}

	machines, err := machines.Load(context.Background(), repo, c.MachineServerVerbosity)
	if err != nil {
		slog.Error("failed to load the machines", "error", err)
		os.Exit(1)
	}
	defer machines.Close()

	inspector, err := inspect.New(machines)
	if err != nil {
		slog.Error("failed to create inspector", "error", err)
		os.Exit(1)
	}
```

I still have to see the advancer initialization code but, perhaps we could move part of this to a separate function responsible for creating the service

internal/inspect/inspect_test.go
```diff
@@ -0,0 +1,242 @@
...
+	s.ServicePort++
+	s.ServiceAddr = fmt.Sprintf("127.0.0.1:%v", s.ServicePort)
+}
+
+func (s *InspectSuite) TestNew() {
+	s.Run("Ok", func() {
+		require := s.Require()
+		machines := newMockMachines()
+		machines.Map[randomAddress()] = &MockMachine{}
+		inspect, err := New(machines)
```
marcelstanley, line 52: `s/inspect/inspector/g`

internal/inspect/inspect.go
```diff
@@ -0,0 +1,156 @@
...
+				"err", "Missing payload")
+			http.Error(w, "Missing payload", http.StatusBadRequest)
+			return
+		}
+		payload, err = hexutil.Decode(r.PathValue("payload"))
+		if err != nil {
+			slog.Info("Internal server error",
+				"service", "inspect",
+				"err", err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
marcelstanley: A failure in decoding would suggest a wrongly encoded payload, right? Something like this:
```suggestion
			http.Error(w, err.Error(), http.StatusBadRequest)
```

internal/inspect/inspect.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	for _, report := range result.Reports {
+		reports = append(reports, hexutil.Encode(report))
+	}
+
+	if result.Accepted {
+		status = "Accepted"
+	} else {
+		status = "Refused"
```
marcelstanley: Perhaps...
```suggestion
		status = "Rejected"
```

internal/inspect/inspect.go
```diff
@@ -0,0 +1,156 @@
...
+var (
+	ErrInvalidMachines = errors.New("machines must not be nil")
+	ErrNoApp           = errors.New("no machine for application")
+)
+
+type Inspector struct {
+	machines Machines
+}
+
+type InspectorResponse struct {
```
marcelstanley:
```suggestion
type InspectResponse struct {
```

internal/inspect/inspect.go
renan061: I think it is safe to break the old design in this case.

internal/inspect/inspect_test.go
GMKrieger: I don't understand the request

marcelstanley: Sorry, please replace `inspect` with `inspector` wherever it seems fit.

mpolitzer (dismissed):

mpolitzer (approved):

PR #585, opened 2024-08-30
fix: fix the CI make-generate to use an image with the cartesi machine

renan061:


.github/workflows/build.yml
```diff
@@ -40,16 +85,38 @@ jobs:
...
+      - build-ci-base
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Fix VCS Issue
+        run : git config --global --add safe.directory /__w/rollups-node/rollups-node
```
fmoura, line 109: We need to correctly generate the "ci-base" image , with the correct user and permissions, so step is not necessary anymore

fmoura (dismissed):

.github/workflows/build.yml
renan061: You mean the `Fix VCS Issue` step?

vfusco (approved):

fmoura (approved):

PR #586, opened 2024-08-30
feat: move snapshot to the tooling folder and add db package 

renan061:


test/tooling/db/db.go
```diff
@@ -0,0 +1,77 @@
...
+	err := SetupSchema(endpoint)
+	if err != nil {
+		return "", err
+	}
+
+	return endpoint, nil
+}
+
+func SetupContainer(ctx context.Context) (*postgres.PostgresContainer, error) {
+	log := "database system is ready to accept connections"
```
fmoura, line 47: It would be better to use our "deps" package here. As in the future we want to use Github Services instead of Testcontainers. "deps" will serve as an intermediate layer. You can start only the Database by disabling Devnet

renan061: Currently, `deps` does too much. I wouldn't like to mix it with this code, that is very self contained.

In the future, we should strip `deps` for parts and create smaller packages (probably in this very */tooling* folder).

vfusco (approved):

fmoura (approved):

vfusco (approved):

PR #587, opened 2024-08-30
Feature/evm reader claim status

fmoura:


internal/evmreader/evmreader.go
```diff
@@ -39,6 +40,12 @@ type EvmReaderRepository interface {
 	GetAllRunningApplications(ctx context.Context) ([]Application, error)
 	GetNodeConfig(ctx context.Context) (*NodePersistentConfig, error)
 	GetEpoch(ctx context.Context, indexKey uint64, appAddressKey Address) (*Epoch, error)
+	GetPreviousSubmittedClaims(ctx context.Context, app Address, lastBlock uint64) ([]Epoch, error)
+	StoreClaimsTransaction(ctx context.Context,
+		app Address,
+		claims []*Epoch,
```
torives, line 47: Calling an `Epoch` struct a claim is bound to cause confusion and trouble in the future. Those are different concepts, even though they currently share a table in the database.

To better accommodate such scenarios we should discuss changes to the `model` package after the alpha release, but right now I'd rather call it what it is and better document the code to clarify what it is doing.

And to be clear, I'm also referring to all other instances of this throughout the code.

internal/evmreader/evmreader.go
```diff
@@ -55,6 +62,8 @@ type EthWsClient interface {
 
 type ConsensusContract interface {
 	GetEpochLength(opts *bind.CallOpts) (*big.Int, error)
+	RetrieveClaimAcceptanceEvents(opts *bind.FilterOpts, appAddresses []common.Address,
+	) ([]*iconsensus.IConsensusClaimAcceptance, error)
```
torives:
```suggestion
	RetrieveClaimAcceptanceEvents(
		opts *bind.FilterOpts, 
		appAddresses []common.Address,
	) ([]*iconsensus.IConsensusClaimAcceptance, error)
```

internal/evmreader/evmreader.go
```diff
@@ -74,6 +83,14 @@ func (e *SubscriptionError) Error() string {
...
 
+// Internal struct to hold application and it's
+// contracts together
+type application struct {
+	Application
+	applicationContract ApplicationContract
+	consensusContract   ConsensusContract
+}
+
 // EvmReader reads inputs from the blockchain
```
torives: It is doing quite a lot more now. I suggest updating its description to document its roles better.

internal/evmreader/evmreader.go
```diff
@@ -147,113 +164,58 @@ func (r *EvmReader) watchForNewBlocks(ctx context.Context, ready chan<- struct{}
...
 		case <-headers:
+
 			// Every time a new block arrives
-			err = r.checkForNewInputs(ctx)
+
+			// Get All Applications
+			runningApps, err := r.repository.GetAllRunningApplications(ctx)
 			if err != nil {
-				slog.Error("Error checking for new inputs",
+				slog.Error("Error retrieving running applications for new inputs",
```
torives:
```suggestion
				slog.Error("Error retrieving running applications",
```

internal/repository/evmreader.go
```diff
@@ -239,3 +242,132 @@ func (pg *Database) GetLastProcessedBlock(
...
+}
+
+func (pg *Database) StoreClaimsTransaction(
+	ctx context.Context,
+	app Address,
+	claims []*Epoch,
+	blockNumber uint64,
+) error {
+
+	var errUpdateClaims = errors.New("unable to update claims")
```
torives: This should be moved to the top of the file, as should all the other custom error declarations.

internal/evmreader/evmreader.go
```diff
@@ -39,6 +40,12 @@ type EvmReaderRepository interface {
 	GetAllRunningApplications(ctx context.Context) ([]Application, error)
 	GetNodeConfig(ctx context.Context) (*NodePersistentConfig, error)
 	GetEpoch(ctx context.Context, indexKey uint64, appAddressKey Address) (*Epoch, error)
+	GetPreviousSubmittedClaims(ctx context.Context, app Address, lastBlock uint64) ([]Epoch, error)
```
marcelstanley:
```suggestion
	GetSubmittedClaims(ctx context.Context, app Address, lastBlock uint64) ([]Epoch, error)
```

internal/evmreader/evmreader.go
```diff
@@ -74,6 +83,14 @@ func (e *SubscriptionError) Error() string {
 	return fmt.Sprintf("Subscription error : %v", e.Cause)
 }
 
+// Internal struct to hold application and it's
+// contracts together
```
marcelstanley:
```suggestion
// Internal struct to hold application and its contracts together
```

internal/evmreader/evmreader.go
```diff
@@ -147,113 +164,58 @@ func (r *EvmReader) watchForNewBlocks(ctx context.Context, ready chan<- struct{}
...
+					consensusContract:   consensusContract})
+			}
 
-			slog.Info("Checking inputs for applications",
-				"apps", appAddresses,
-				"last processed block", lastProcessedBlock,
-				"most recent block", mostRecentBlockNumber,
-			)
+			if len(apps) == 0 {
+				slog.Info("No running consistent applications")
```
marcelstanley: Is this a typo?
```suggestion
				slog.Info("No running applications")
```

internal/evmreader/evmreader.go
```diff
@@ -288,235 +250,46 @@ func (r *EvmReader) fetchMostRecentHeader(
...
 
 	}
-	return consensus, nil
+	return applicationContract, consensus, nil
 }
 
-// Reads the application epoch length given it's consesus contract
-func (r *EvmReader) getEpochLengthFromContract(consensus ConsensusContract) (uint64, error) {
+// getEpochLengthFromContract reads the application epoch length given it's consensus contract
+func getEpochLengthFromContract(consensus ConsensusContract) (uint64, error) {
```
marcelstanley:
```suggestion
func getEpochLength(consensus ConsensusContract) (uint64, error) {
```

internal/evmreader/evmreader.go
```diff
@@ -529,73 +302,50 @@ func (r *EvmReader) getEpochLengthFromContract(consensus ConsensusContract) (uin
...
 
-// Calculates the epoch index given the input block number
+// calculateEpochIndex calculates the epoch index given the input block number
+// and epoch length
 func calculateEpochIndex(epochLength uint64, blockNumber uint64) uint64 {
 	return blockNumber / epochLength
 }
 
-func appToAddresses(apps []Application) []Address {
+func appToAddresses(apps []application) []Address {
```
marcelstanley:
```suggestion
func appsToAddresses(apps []application) []Address {
```

internal/evmreader/evmreader.go
```diff
@@ -529,73 +302,50 @@ func (r *EvmReader) getEpochLengthFromContract(consensus ConsensusContract) (uin
...
 	var addresses []Address
 	for _, app := range apps {
 		addresses = append(addresses, app.ContractAddress)
 	}
 	return addresses
 }
 
+// sortByInputIndex is a compare function that orders Inputs
+// by it's index field. It is intended to be used with
```
marcelstanley:
```suggestion
// by index. It is intended to be used with
```

internal/repository/evmreader.go
```diff
@@ -239,3 +242,132 @@ func (pg *Database) GetLastProcessedBlock(
...
+			"id":     claim.Id,
+		}
+
+		tag, err := tx.Exec(ctx, updateEpochQuery, updateClaimArgs)
+		if err != nil {
+			return errors.Join(errUpdateClaims, err, tx.Rollback(ctx))
+		}
+		if tag.RowsAffected() != 1 {
+			return errors.Join(errUpdateClaims,
+				fmt.Errorf("no rows affected when updating claim %d", claim.Index),
```
torives: With this check, you can only affirm with certainty that your expectation of a single affected row wasn't met.

internal/evmreader/evmreader.go
```diff
@@ -529,73 +302,50 @@ func (r *EvmReader) getEpochLengthFromContract(consensus ConsensusContract) (uin
...
-			AppAddress:       event.AppContract,
-		}
-
-		// Insert Sorted
-		appInputsMap[event.AppContract] = insertSorted(appInputsMap[event.AppContract], input)
-	}
-	return appInputsMap, nil
-}
-
 // Util functions
```
marcelstanley: These util functions should be moved to its own file as they are only used elsewhere, at `claim.go` and `input.go`

internal/evmreader/consensus_adapter.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	defer itr.Close()
+
+	var events []*iconsensus.IConsensusClaimAcceptance
+	for itr.Next() {
+		claimAcceptanceEvent := itr.Event
+		events = append(events, claimAcceptanceEvent)
+	}
+	err = itr.Error()
+	if err != nil {
```
torives:
```suggestion
	if err = itr.Error(); err != nil {
```

internal/evmreader/evmreader.go
torives: Is this a typo?

internal/evmreader/evmreader.go
```diff
@@ -147,113 +164,58 @@ func (r *EvmReader) watchForNewBlocks(ctx context.Context, ready chan<- struct{}
...
+					applicationContract: applicationContract,
+					consensusContract:   consensusContract})
+			}
 
-			slog.Info("Checking inputs for applications",
-				"apps", appAddresses,
-				"last processed block", lastProcessedBlock,
-				"most recent block", mostRecentBlockNumber,
-			)
+			if len(apps) == 0 {
```
torives, line 195: IMO checking this right after the repository call makes more sense.

internal/evmreader/evmreader.go
torives: This function is only called by the code in the `input.go` file. I suggest moving it there.

internal/evmreader/retrypolicy/retrypolicy_consensus_delegator.go
```diff
@@ -33,27 +35,37 @@ func NewConsensusWithRetryPolicy(
...
-	args getEpochLengthArgs,
-) (*big.Int, error) {
-	return d.delegate.GetEpochLength(args.opts)
+type retrieveClaimAcceptedEventsArgs struct {
+	opts         *bind.FilterOpts
+	appAddresses []common.Address
+}
+
+func (d *ConsensusRetryPolicyDelegator) RetrieveClaimAcceptanceEvents(
+	opts *bind.FilterOpts, appAddresses []common.Address,
```
torives:
```suggestion
	opts *bind.FilterOpts, 
	appAddresses []common.Address,
```

internal/evmreader/claim.go
```diff
@@ -0,0 +1,220 @@
...
+	// that can happen before claim submission
+
+	// Classify them by same IConsensusAddress
+	sameConsensusApps := indexApps(keyByIConsensus, apps)
+	for iConsensusAddress, apps := range sameConsensusApps {
+
+		appAddresses := appToAddresses(apps)
+
+		// All these apps shares the same IConsensus
+		if len(apps) == 0 {
```
torives: This makes no sense. If there are no apps for an IConsensus address, this address wouldn't even be a key in the `sameConsensusApps` map.

internal/evmreader/claim.go
```diff
@@ -0,0 +1,220 @@
...
+						r.epochLengthCache[app],
+						claimAcceptance.LastProcessedBlockNumber.Uint64()),
+					app)
+				if err != nil {
+					slog.Error("Error retrieving claim", "app", app, "error", err)
+				}
+
+				// Check Claim
+				if claim == nil {
+					slog.Error("Got unknown claim event",
```
torives: This claim could be nil because the repository failed to retrieve the epoch due to a network issue, for example. Then your error reason would be incorrect.

internal/evmreader/claim.go
```diff
@@ -0,0 +1,220 @@
...
+
+				// Get  Claim
+				claim, err := r.repository.GetEpoch(
+					ctx, calculateEpochIndex(
+						r.epochLengthCache[app],
+						claimAcceptance.LastProcessedBlockNumber.Uint64()),
+					app)
+				if err != nil {
+					slog.Error("Error retrieving claim", "app", app, "error", err)
+				}
```
torives: Could you document this logic better? I don't fully understand the consequences of the code paths where errors occur but the function/loop execution continues.

internal/evmreader/claim.go
fmoura: I just got cautious, in case the `indexApps` function had a bug

internal/evmreader/claim.go
fmoura: The error before was not correctly handled. Corrected it. The only way the claim is nil, is being absent from the DB

internal/evmreader/evmreader.go
fmoura: Applications could be left out because of IConsensus inconsistencies, since the blockchain registered IConsensus could be different from the DB registered one.
We should process only the "consistent" ones

internal/evmreader/input.go
```diff
@@ -0,0 +1,322 @@
...
+		}
+
+		// Insert Sorted
+		appInputsMap[event.AppContract] = insertSorted(
+			sortByInputIndex, appInputsMap[event.AppContract], input)
+	}
+	return appInputsMap, nil
+}
+
+// // byLastProcessedBlock key extractor function intended to be used with `indexApps` function
```
GMKrieger:
```suggestion
// byLastProcessedBlock key extractor function intended to be used with `indexApps` function
```

internal/repository/evmreader.go
```diff
@@ -239,3 +245,130 @@ func (pg *Database) GetLastProcessedBlock(
...
+	blockNumber uint64,
+) error {
+
+	updateEpochQuery := `
+	UPDATE epoch
+	SET
+		status = @status
+	WHERE
+		id = @id
+	`
```
GMKrieger:
```suggestion
		id = @id`
```

internal/repository/evmreader.go
```diff
@@ -239,3 +245,130 @@ func (pg *Database) GetLastProcessedBlock(
...
+				tx.Rollback(ctx))
+		}
+	}
+
+	// Update last processed block
+	updateLastBlockQuery := `
+	UPDATE application
+	SET last_claim_check_block = @blockNumber
+	WHERE
+		contract_address=@contractAddress`
```
GMKrieger:
```suggestion
	UPDATE application
	SET
	        last_claim_check_block = @blockNumber
	WHERE
		contract_address=@contractAddress`
```

GMKrieger (commented):
Just some minor formatting details. Also, you need to squash the fixups

internal/evmreader/retrypolicy/contratfactory.go
```diff
@@ -7,17 +7,17 @@ import (
```
marcelstanley, line 7: There's a typo in the file name. Please rename it to "contra**c**tfactory.go".

internal/evmreader/claim.go
torives: Then please add unit tests for it and remove this check. The simpler, the better.

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+	ctx context.Context,
+	apps []application,
+	lastClaimCheck, mostRecentBlockNumber uint64,
+) {
+
+	// DISCLAIMER: The current algorithm will only handle Authority.
+	// To handle Quorum, node needs to handle acceptance events
+	// that can happen before claim submission
+
+	// Classify them by same IConsensusAddress
```
torives:
```suggestion
	// Classify them by the same IConsensusAddress
```

internal/repository/validator.go
```diff
@@ -23,7 +23,7 @@ func (pg *Database) GetOutputsProducedInBlockRange(
 	endBlock uint64,
 ) ([]Output, error) {
 	query := `
-	SELECT
```
marcelstanley, line 26: Formatting changes should be moved to a separate chore commit.

internal/evmreader/claim.go
```diff
@@ -98,6 +98,7 @@ func (r *EvmReader) readAndUpdateClaims(
 		}
 
 		// Check events against Epochs
+	APP_LOOP:
```
torives, line 100: This label is unnecessary. The added `continue` statements are enough.

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+	"context"
+	"log/slog"
+
+	. "github.com/cartesi/rollups-node/internal/node/model"
+	"github.com/cartesi/rollups-node/pkg/contracts/iconsensus"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+func (r *EvmReader) checkForClaimStatus(
```
torives, line 17:
```suggestion
func (r *EvmReader) updateClaimsStatus(
```

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+					slog.Error("Got claim acceptance event for an unknown epoch",
+						"app", app,
+						"claim last block", claimAcceptance.LastProcessedBlockNumber,
+						"hash", claimAcceptance.Claim)
+					continue APP_LOOP
+				}
+
+				// Update Epoch claim status
+				if claimAcceptance.Claim != *epoch.ClaimHash ||
+					claimAcceptance.LastProcessedBlockNumber.Uint64() != epoch.LastBlock {
```
torives, line 147: If this line is true, doesn't it mean we have an inconsistent state? 

We use `claimAcceptance.LastProcessedBlockNumber` to calculate the epoch index. If it does not match with `epoch.LastBlock`, it means that `calculateEpochIndex` is wrong or that the epoch record in the database is incorrect.

It is safe to assume this is not the case, and this second check can be removed. But we can discuss this further if necessary.

internal/repository/evmreader.go
```diff
@@ -239,3 +245,130 @@ func (pg *Database) GetLastProcessedBlock(
 
 	return block, nil
 }
+
+func (pg *Database) GetPreviousSubmittedClaimsEpochs(
```
marcelstanley: Please add some documentation to the function and rename it to
```suggestion
func (pg *Database) GetClaimSubmittedEpochs(
```

and fix the name used in the error messages below. Example:
https://github.com/cartesi/rollups-node/pull/587/files?diff=unified&w=0#diff-e4c6cf8eaabd1c2b1101fd6881c35efcd8ab6f4a07bbe52148cd947d80335c1fR279

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	claimAcceptanceEvents, err := consensusContract.RetrieveClaimAcceptanceEvents(
+		opts, appAddresses)
+	if err != nil {
+		return nil, err
+	}
+	for _, event := range claimAcceptanceEvents {
+		appClaimAcceptanceMap[event.AppContract] = insertSorted(
+			sortByLastBlockNumber, appClaimAcceptanceMap[event.AppContract], event)
+	}
```
torives, line 200: You can remove the extra loop by checking the map contents before inserting, with the added benefit of removing from the map addresses of apps with no events.

```suggestion
	opts := &bind.FilterOpts{
		Context: ctx,
		Start:   startBlock,
		End:     &endBlock,
	}
	claimAcceptanceEvents, err := consensusContract.RetrieveClaimAcceptanceEvents(
		opts, appAddresses)
	if err != nil {
		return nil, err
	}
	for _, event := range claimAcceptanceEvents {
		if _, ok := appClaimAcceptanceMap[event.AppContract]; ok {
			appClaimAcceptanceMap[event.AppContract] = insertSorted(
				sortByLastBlockNumber,
				appClaimAcceptanceMap[event.AppContract],
				event,
			)
		} else {
			appClaimAcceptanceMap[event.AppContract] = insertSorted(
				sortByLastBlockNumber,
				[]*iconsensus.IConsensusClaimAcceptance{event},
				event,
			)
		}
	}
```

internal/repository/base_test.go
```diff
@@ -120,6 +120,20 @@ func (s *RepositorySuite) SetupDatabase() {
 	_, err = s.database.InsertEpoch(s.ctx, &epoch2)
 	s.Require().Nil(err)
 
+	epoch3 := Epoch{
```
marcelstanley, line 123: We should partition the tests better. Having `RepositorySuite` being shared by different test files is a bit cumbersome.

internal/repository/evmreader.go
```diff
@@ -239,3 +245,130 @@ func (pg *Database) GetLastProcessedBlock(
...
+	if err != nil {
+		return nil, fmt.Errorf("GetProcessedEpochs failed: %w", err)
+	}
+	return results, nil
+}
+
+func (pg *Database) UpdateEpochsClaimStatusTransaction(
+	ctx context.Context,
+	app Address,
+	claims []*Epoch,
```
marcelstanley: I'd suggest adding some documentation to the function.

Also:
```suggestion
func (pg *Database) UpdateEpochsStatus(
	ctx context.Context,
	app Address,
	epochs []*Epoch,
```

internal/repository/evmreader.go
```diff
@@ -239,3 +245,130 @@ func (pg *Database) GetLastProcessedBlock(
...
+		updateClaimArgs := pgx.NamedArgs{
+			"status": claim.Status,
+			"id":     claim.Id,
+		}
+
+		tag, err := tx.Exec(ctx, updateEpochQuery, updateClaimArgs)
+		if err != nil {
+			return errors.Join(errUpdateEpochs, err, tx.Rollback(ctx))
+		}
+		if tag.RowsAffected() != 1 {
```
marcelstanley, line 346: Is this possible?

internal/evmreader/evmreader.go
marcelstanley: +1

internal/evmreader/evmreader.go
```diff
@@ -25,7 +26,7 @@ import (
 type InputSource interface {
 	// Wrapper for FilterInputAdded(), which is automatically generated
 	// by go-ethereum and cannot be used for testing
-	RetrieveInputs(opts *bind.FilterOpts, appContract []common.Address, index []*big.Int,
+	RetrieveInputs(opts *bind.FilterOpts, appAddresses []common.Address, index []*big.Int,
```
torives: Please use the type alias from the `model` package. The same goes for the other files.

```suggestion
	RetrieveInputs(opts *bind.FilterOpts, appAddresses []Address, index []*big.Int,
```

internal/evmreader/consensus_adapter.go
```diff
@@ -0,0 +1,57 @@
...
+	"math/big"
+
+	"github.com/cartesi/rollups-node/pkg/contracts/iconsensus"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+// IConsensus Wrapper
+type ConsensusContractAdapter struct {
```
marcelstanley, line 16: `s/ConsensusContractAdapter/ConsensusAdapter/g`

internal/evmreader/retrypolicy/contratfactory.go
```diff
@@ -52,12 +53,19 @@ func (f *EvmReaderContractFactory) NewIConsensus(
...
 
-	// Building a contract does not fail due to network errors.
-	// No need to retry this operation
-	consensusContract, err := iconsensus.NewIConsensus(address, f.ethClient)
-	if err != nil {
-		return nil, err
+	delegator, ok := f.iConsensusCache[address]
+	if !ok {
+		// Building a contract does not fail due to network errors.
+		// No need to retry this operation
```
marcelstanley, line 59: It may fail on [`FilterClaimAcceptance()`](https://github.com/cartesi/rollups-node/blob/92a0535ddc0a05261aa50ff785b80dd686b70ec0/pkg/contracts/iconsensus/iconsensus.go#L344) , right?
```suggestion
```

internal/evmreader/input.go
```diff
@@ -0,0 +1,322 @@
...
+func byLastProcessedBlock(app application) uint64 {
+	return app.LastProcessedBlock
+}
+
+// getEpochLengthFromContract reads the application epoch length given it's consensus contract
+func getEpochLengthFromContract(consensus ConsensusContract) (uint64, error) {
```
marcelstanley:
```suggestion
// getEpochLength reads the epoch length from a consensus 
func getEpochLength(consensus ConsensusContract) (uint64, error) {
```

internal/evmreader/evmreader.go
```diff
@@ -39,6 +40,16 @@ type EvmReaderRepository interface {
 	GetAllRunningApplications(ctx context.Context) ([]Application, error)
 	GetNodeConfig(ctx context.Context) (*NodePersistentConfig, error)
 	GetEpoch(ctx context.Context, indexKey uint64, appAddressKey Address) (*Epoch, error)
+	GetPreviousSubmittedClaimsEpochs(
```
torives: The current name is quite odd. I would rather choose something like `GetEpochsWithOpenClaims` or `GetEpochsWithUnfinishedClaims`, as "previously submitted claims" is an ambiguous term that includes approved or rejected claims.

internal/evmreader/evmreader.go
```diff
@@ -128,7 +151,7 @@ func (r *EvmReader) Run(ctx context.Context, ready chan<- struct{}) error {
 	}
 }
 
-// Watch for new blocks and reads new inputs based on the
+// watchForNewBlocks watches for new blocks and reads new inputs based on the
 // default block configuration, which have not been processed yet.
```
marcelstanley, line 151:
```suggestion
// watchForNewBlocks watches for new blocks and reads InputBoxInputAdded and ClaimAcceptance
// events, which have not been processed yet, based on the default block configuration.
```

internal/evmreader/evmreader.go
```diff
@@ -74,7 +89,15 @@ func (e *SubscriptionError) Error() string {
 	return fmt.Sprintf("Subscription error : %v", e.Cause)
 }
 
-// EvmReader reads inputs from the blockchain
+// Internal struct to hold application and it's
+// contracts together
+type application struct {
+	Application
+	applicationContract ApplicationContract
+	consensusContract   ConsensusContract
```
marcelstanley, line 93: We need better names for both

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+	// To handle Quorum, node needs to handle acceptance events
+	// that can happen before claim submission
+
+	// Classify them by same IConsensusAddress
+	sameConsensusApps := indexApps(keyByIConsensus, apps)
+	for iConsensusAddress, apps := range sameConsensusApps {
+
+		appAddresses := appToAddresses(apps)
+
+		// All these apps shares the same IConsensus
```
marcelstanley:
```suggestion
		// All apps share the same IConsensus
```

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+			epochs := []*Epoch{}
+			for _, claimAcceptance := range claimAcceptances {
+
+				// Get Previous Epochs with submitted claims, that got no acceptance event
+				// and update their statuses rejecting them all
+				previousEpochs, err := r.repository.GetPreviousSubmittedClaimsEpochs(
+					ctx, app, claimAcceptance.LastProcessedBlockNumber.Uint64())
+				if err != nil {
+					slog.Error("Error retrieving previous submitted claims",
+						"app", app, "error", err)
```
marcelstanley:
```suggestion
						"app", app,
						"block", claimAcceptance.LastProcessedBlockNumber.Uint64(),
						 "error", err)
```

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+				}
+
+				// Get the Epoch for the current Claim Acceptance Event
+				epoch, err := r.repository.GetEpoch(
+					ctx, calculateEpochIndex(
+						r.epochLengthCache[app],
+						claimAcceptance.LastProcessedBlockNumber.Uint64()),
+					app)
+				if err != nil {
+					slog.Error("Error retrieving Epoch", "app", app, "error", err)
```
marcelstanley:
```suggestion
					slog.Error("Error retrieving Epoch",
						"app", app,
						"block", claimAcceptance.LastProcessedBlockNumber.Uint64(),
						 "error", err)
```

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+				previousEpochs, err := r.repository.GetPreviousSubmittedClaimsEpochs(
+					ctx, app, claimAcceptance.LastProcessedBlockNumber.Uint64())
+				if err != nil {
+					slog.Error("Error retrieving previous submitted claims",
+						"app", app, "error", err)
+					continue APP_LOOP
+				}
+
+				for _, previousEpoch := range previousEpochs {
+					previousEpoch.Status = EpochStatusClaimRejected
```
marcelstanley: The fact there was no acceptance for these epochs is unexpected. We should stop processing here, right?.

@fmoura as we discussed, should we keep reading the blockchain for such events for the current app? How should we signal this sort of failure to the node operator?

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+						r.epochLengthCache[app],
+						claimAcceptance.LastProcessedBlockNumber.Uint64()),
+					app)
+				if err != nil {
+					slog.Error("Error retrieving Epoch", "app", app, "error", err)
+					continue APP_LOOP
+				}
+
+				// Check Epoch
+				if epoch == nil {
```
marcelstanley, line 138: This is another unexpected scenario

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+				continue
+			}
+
+		}
+
+	}
+
+}
+
+func (r *EvmReader) readClaimAcceptanceFromBlockchain(
```
marcelstanley:
```suggestion
func (r *EvmReader) readClaimAcceptances(
```

internal/evmreader/claim.go
```diff
@@ -0,0 +1,224 @@
...
+						"app", app,
+						"claim last block", claimAcceptance.LastProcessedBlockNumber,
+						"hash", claimAcceptance.Claim)
+					continue APP_LOOP
+				}
+
+				// Update Epoch claim status
+				if claimAcceptance.Claim != *epoch.ClaimHash ||
+					claimAcceptance.LastProcessedBlockNumber.Uint64() != epoch.LastBlock {
+					slog.Warn("Claim Rejected",
```
marcelstanley: Another unexpected scenario that should be signaled properly
```suggestion
					slog.Warn("Claim Mismatch",
```

marcelstanley (dismissed):
If possible, split the changes into separate commits, specially because of the refactoring you did, @fmoura

internal/evmreader/claim.go
fmoura: Yeah. Updating it. It actually, when handling Authority, is an indication of the Application reaching an invalid state. It will be handled at #594

internal/evmreader/claim.go
fmoura: We have two loops , the outer is a "for each Application", and the inner a "for each Claim Acceptance Event for this Application".

If an error occurs we stop processing the current Application, so we want to continue the outer loop, not the inner `Claim Acceptance Event` one.

internal/evmreader/claim.go
fmoura: Even applications with no Acceptance Events have it's `LastClaimCheckBlock`updated at each run. That's why they are in the map

internal/evmreader/claim.go
fmoura: This will be handled here #594

internal/repository/base_test.go
fmoura: I prefer doing this on another PR . One taking care of this and the evm-reader folder re-organization

internal/evmreader/evmreader.go
fmoura: Went with @torives suggestion above

internal/evmreader/claim.go
torives: I overlooked the `for` loop for each claim event, my bad. By the way, it wouldn't hurt to extract one or more functions from this code. We have four nested loops in here, it's hard to keep track. 

My two candidates would be the code handling previous epochs and the one handling a single `claimAcceptance` event.

internal/repository/evmreader.go
fmoura: It is unlikely, but possible, as if is there some inconsistency in the data

internal/evmreader/retrypolicy/contratfactory.go
fmoura: Yeah, it can fail later. But retry policy is taking care of those

vfusco (approved):
Approving as is with the intent of fixing the implementation later

PR #591, opened 2024-09-03
fix(db): Fix Schema postgres ssl mode handling

fmoura:


internal/node/startup/startup.go
```diff
@@ -22,7 +22,7 @@ import (
 // Validates the Node Database Schema Version
 func ValidateSchema(config config.NodeConfig) error {
 	endpoint := config.PostgresEndpoint.Value
-	if config.PostgresSslMode {
+	if !config.PostgresSslMode {
```
vfusco: Why don't we call this disable ssl?

vfusco (changes requested):

internal/node/startup/startup.go
fmoura: Our current policy is to have the configs always as "enable feature X"  to avoid user confusion. We used to have a mixture of "feature A enable" and "feature B disable" properties.

marcelstanley: +1 for `disableSsl`. This could be the start of the normalization of all boolean configuration parameters we have.

GMKrieger (dismissed):

internal/node/config/generate/Config.toml
```diff
@@ -206,11 +206,11 @@ It is also possible to set the endpoint without a password and load it from Post
 See [this](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-PASSFILE)
 for more information."""
 
-[postgres.CARTESI_POSTGRES_SSLMODE_ENABLED]
-default = "true"
+[postgres.CARTESI_POSTGRES_SSL_MODE_DISABLE]
+default = "false"
```
marcelstanley: Let's simplify it for now, so we can move forward
```suggestion
[postgres.CARTESI_SSL_DISABLED]
default = "false"
```

We can revisit it later on

marcelstanley (approved):

internal/node/startup/startup.go
renan061: @marcelstanley  I don't understand. We already normalized. Everything is SomethingEnabled, as @fmoura pointed out.

I don't mind changing the NodeConfig field to *disable* so much, but I don't see the benefit.

GMKrieger (approved):
I'm approving because it's working and we need to get things moving, but for 2.0 proper we need to have a discussion about config pattern on the Node because getting the not value of something is going to cause headaches in the future.

vfusco (approved):

PR #593, opened 2024-09-03
Feature/evm reader output executed

fmoura:


vfusco (approved):
Approving as is with the intent of refactoring and fixing it later

PR #603, opened 2024-10-15
feat: Create a template service

mpolitzer:
Template of a service with:
 - cancellation
 - liveliness check
 - readiness check
 - logging
 - polling
 - reloading
 - signal handling


mpolitzer:
moved: https://github.com/cartesi/rollups-node/pull/605

PR #615, opened 2025-01-20
Feature/new build makefile gitignore

miltonjonat:
Proposes two small chore changes:
1. Add quotes to the `PATH` env var in Makefile, to prevent issues should the path contain spaces

2. Add the `applications` directory to .gitignore, to avoid including app snapshots such as the one for `echo-dapp`


vfusco:
This is already fixed.

PR #617, opened 2025-01-24
feat(claimer): find claim submission event in chunks

mpolitzer:
- `chunkedFindClaimSubmissionEvent` retrives claim submission events from the
blockchain via the eth_getLogs calls down the stack. However depending on
the provider, there may be limitations on the block range, number of events,
payload size... etc returned and a single query may be "too heavy" and fail.
To circumvent this issue, we'll progressively split the query into multiple
queries instead in case a larger query fails.

- Convert `FindClaimSubmissionEventAndSucc` loop to pull mode.


mpolitzer:
migrated to: #623

PR #631, opened 2025-03-26
Feature/add get processed inputs and last accepted epoch

mpolitzer:


internal/jsonrpc/jsonrpc-discover.json
```diff
@@ -138,6 +138,27 @@
 				}
 			}
 		},
+		{
+			"name": "cartesi_getLastAcceptedEpoch",
+			"summary": "Get the last accepted epoch",
+			"description": "Fetches a single epoch by application and index.",
```
tuler: No index

internal/jsonrpc/jsonrpc.go
```diff
@@ -465,23 +503,22 @@ func (s *Service) handleGetProcessedInputCount(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	app, err := s.repository.GetApplication(r.Context(), params.Application)
+	processedInputs, err := s.repository.GetProcessedInputs(r.Context(), params.Application)
+	if errors.Is(err, sql.ErrNoRows) {
```
vfusco: Please don't leak sql.* errors. Create a proper respository error. eg ApplicationNotFound

vfusco (changes requested):

internal/jsonrpc/jsonrpc.go
mpolitzer: ok, PR updated

vfusco (approved):

PR #632, opened 2025-03-27
Update machine emulator to v0.19.0

vfusco:


pkg/emulator/machine.go
```diff
@@ -1,250 +1,267 @@
...
+	var cCmd C.uint8_t
+	var cReason C.uint16_t
+
+	// We'll guess some large buffer.
+	// If the new API pattern is different, adapt as needed.
+	maxBuf := 2 * 1024 * 1024
+	buf := make([]byte, maxBuf)
+	var cLen C.uint64_t = C.uint64_t(maxBuf)
+
+	e := newError(C.cm_receive_cmio_request(
```
mpolitzer, line 214: We can use this pattern to retrieve the exact size, allocate a buffer and then read the data.
(Can be left as a future improvement. I'll just mark this as resolved here)
```
    uint64_t length{};
    if (cm_receive_cmio_request(m.get(), nullptr, nullptr, nullptr, &length) != 0) {
        return luaL_error(L, "%s", cm_get_last_error_message());
    uint16_t reason{};
    if (cm_receive_cmio_request(m.get(), &cmd, &reason, data, &length) != 0) {
        return luaL_error(L, "%s", cm_get_last_error_message());
```

mpolitzer (approved):

PR #642, opened 2025-04-07
fix: validator when there are no outputs

mpolitzer:


internal/validator/validator.go
```diff
@@ -154,14 +154,17 @@ func (v *Service) validateApplication(ctx context.Context, app *Application) err
...
+			v.Logger.Error("failed to create claim and proofs.", "error", err)
+			return err
+		}
+
 		v.Logger.Info("Claim Computed",
 			"application", appAddress,
 			"epoch_index", epoch.Index,
-			"last_block", epoch.LastBlock,
+			"claimHash", *claim,
+			"error", err,
```
endersonmaia: This will always be `nil`, I'd remove this line.

```suggestion
```

vfusco (approved):

PR #653, opened 2025-04-22
getLastAcceptedEpochIndex

tuler:
Replaces `cartesi_getLastAcceptedEpoch` with `cartesi_getLastAcceptedEpochIndex` that returns only the epoch index, and not the entire `Epoch` object.
I think this is more consistent with the `cartesi_getProcessedInputCount` method, that only returns a number (index), instead of a `cartesi_getLastProcessedInput` that would return an `Input` object.

Typically an application will use the last accepted epoch index to compare with the epoch index of outputs, to know if those outputs can be executed or not, with a "lower than" comparison. A full epoch object is not needed in that case.

I did not change the repository implementation to query only for the index, it is still querying for the epoch object.

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	ErrApplicationNotFound = fmt.Errorf("application not found")
+	ErrNotFound = fmt.Errorf("not found")
 )
 
 type Pagination struct {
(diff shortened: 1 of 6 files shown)


internal/repository/repository.go
```diff
@@ -15,6 +15,7 @@ import (
 
 var (
 	ErrApplicationNotFound = fmt.Errorf("application not found")
+	ErrEpochNotFound = fmt.Errorf("epoch not found")
```
vfusco: I guess we could have just one ErrNotFound

vfusco (changes requested):

internal/repository/repository.go
mpolitzer: updated

vfusco (approved):

PR #655, opened 2025-05-13
fix(ethutil): add alchemy free tier rpc specific response

mpolitzer:
Providers limit calls to `eth_getLogs`, specially in the free tiers. The rpc error code for it is usually not documented can vary wildly.

### DIFF
--- a/pkg/ethutil/filter.go
+++ b/pkg/ethutil/filter.go
@@ -19,19 +19,25 @@ var (
 
 // read chunkedFilterLogs comment for additional information.
 //
-// NOTE: There is no standard reply among providers, add as needed. This
-// function assumes that any server side error codes represent block range that
-// is too large.
-// ┌────────────────────────────┬───────┬────────┬────────────┐
-// │          provider          │ limit │  code  │ checked at │
-// ├────────────────────────────┼───────┼────────┼────────────┤
-// │ https://cloudflare-eth.com │   800 │ -32047 │ 2025-01-24 │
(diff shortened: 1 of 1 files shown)


miltonjonat:
I think it's nice to have these pre-built limits, but wouldn't it also make sense to have a configuration option to manually define the error code for a blockRangeTooLarge error? Alternatively, maybe we could explicitly define the max block range allowed? This way the Node would still be usable should RPC behaviors change - or if the user has an RPC provider we don't explicitly support.

As a concrete example, I get this response testing this now with an excessively large block range on Alchemy: 
```json
{
    "code": -32600,
    "message": "You can make eth_getLogs requests with up to a 500 block range. Based on your parameters, this block range should work: [0x7e9b85, 0x7e9d78]"
}
```

mpolitzer:
> I think it's nice to have these pre-built limits, but wouldn't it also make sense to have a configuration option to manually define the error code for a blockRangeTooLarge error?

I would like to avoid exposing this if possible. But I'll discuss this with Fusco when he gets back from OOO.

> Alternatively, maybe we could explicitly define the max block range allowed?

This function is used in an algorithm to decide if it has to split the query into smaller parts or not. So no need to specify the actual range limits as long as it is not too small (quicknode has a limit of 5).

> As a concrete example, I get this response testing this now with an excessively large block range on Alchemy:

Alchemy should be handled now, did get the error running with this patch?

vfusco:
We will add a max block range config on another PR.

vfusco (approved):

PR #658, opened 2025-05-19
Add Snapshot support and refactor machine runtime

vfusco:


internal/manager/manager.go
```diff
@@ -198,3 +277,22 @@ func getProcessedInputs(ctx context.Context, repo MachineRepository, appAddress
 	f := repository.InputFilter{NotStatus: Pointer(InputCompletionStatus_None)}
 	return repo.ListInputs(ctx, appAddress, f, repository.Pagination{})
 }
+
+// Helper function to get inputs after a specific index
+func getInputsAfterSnapshot(ctx context.Context, repo MachineRepository, app *Application, snapshotInputIndex uint64) ([]*Input, error) {
```
mpolitzer, line 282: You could also take the slice after finding the index. Probably doesn't make much difference.

```
inputs, _, err := getProcessedInputs(ctx, repo, app.IApplicationAddress.String())
if err != nil {
	return nil, err
}

for i, input := range inputs {
	if input.Index > snapshotInputIndex {
		return inputs[i:], nil
	}
}
return []*Input{}, nil
```

pkg/machine/implementation.go
```diff
@@ -0,0 +1,563 @@
...
+	cycle, err := m.server.ReadReg(emulator.REG_MCYCLE)
+	if err != nil {
+		err = fmt.Errorf("could not read the machine's current cycle: %w", err)
+		return cycle, errors.Join(ErrMachineInternal, err)
+	}
+	return cycle, nil
+}
+
+// payloadLengthLimit returns the maximum payload length
+func (m *machineImpl) payloadLengthLimit() uint {
```
mpolitzer:
```
return 1 << emulator.CmioRxBufferLog2Size
```

pkg/machine/implementation.go
```diff
@@ -0,0 +1,563 @@
...
+			}
+			outputs = append(outputs, output)
+		case emulator.AutomaticYieldReasonReport:
+			report, err := m.readMemory(ctx)
+			if err != nil {
+				return outputs, reports, fmt.Errorf("could not read the report: %w", err)
+			}
+			reports = append(reports, report)
+		default:
+			panic("unreachable code: invalid automatic yield reason")
```
mpolitzer: Ignore instead of panic?

pkg/machine/implementation.go
```diff
@@ -0,0 +1,563 @@
...
+	case emulator.BreakReasonReachedTargetMcycle:
+		return nil, currentCycle, nil // returns with no yield type
+	case emulator.BreakReasonHalted:
+		return nil, 0, ErrHalted
+	case emulator.BreakReasonYieldedSoftly:
+		return nil, 0, ErrSoftYield
+	case emulator.BreakReasonFailed:
+		fallthrough // covered by server.Run()
+	default:
+		panic("unreachable code: invalid break reason")
```
mpolitzer: Ignore instead of panic?

pkg/machine/machine_test.go
```diff
@@ -0,0 +1,356 @@
...
+
+func (m *MockMachine) Close(_ context.Context) error {
+	return m.CloseError
+}
+
+func (m *MockMachine) Address() string {
+	return m.AddressReturn
+}
+
+// Test IO functionality
```
mpolitzer: Looks the same as `io_test.go`, can we consolidate?

mpolitzer (commented):
LGTM.

Feel free to incorporate comments at your discretion.

mpolitzer (approved):

pkg/machine/implementation.go
mpolitzer: refactor goes in another PR

pkg/machine/implementation.go
mpolitzer: refactor goes in another PR

pkg/machine/implementation.go
mpolitzer: refactor goes in another PR

pkg/machine/machine_test.go
mpolitzer: refactor goes in another PR

PR #661, opened 2025-05-29
Fix/linter warnings

mpolitzer:


mpolitzer:
this PR is too old, and the rebase is not worth it.

PR #687, opened 2025-07-14
feat(evm-reader): fast sync the last input check block when the appli…

### LINKED ISSUE #686: Speed up first sync of the EVM-reader for new applications
### Context / Problem  
At first startup the EVM-reader issues an `eth_getLogs` **filter-log scan** from the block where the **InputBox** contract was deployed all the way to the latest L1 block.  
For a **brand-new dApp** this huge `fromBlock … toBlock` range contains **hundreds of thousands of empty blocks**, so the node burns JSON-RPC quota and spends minutes waiting for results even though no relevant logs exist.

---

### Suggested Solution  

1. **Fast “probe-and-decide” flow**  
   *Let*  
   `current` = latest L1 block,  
   `appDeploy` = dApp contract deployment block,  
   `inputDeploy` = InputBox deployment block.  

   **Step A – probe `current`**  
   * Call `eth_getLogs` for this dApp **only at `current`**.  
     * **No inputs:** set `last_checked_block = current` and finish (zero-cost start).  
     * **Some inputs:** continue to Step B.  

   **Step B – probe `appDeploy − 1`**  
   * Call `eth_getLogs` at block `appDeploy − 1`.  
     * **No inputs:** sync the shorter range **`[appDeploy … current]`**.  
     * **Some inputs:** fall back to the full scan **`[inputDeploy … current]`**.  

   The optimisation costs at most **two extra RPC calls**.

2. **State & restart safety**  
   Persist `last_checked_block` in the DB so restarts reuse the determined range.

---

### Acceptance Criteria  

| # | Scenario                              | Expected result                                                |
|---|---------------------------------------|----------------------------------------------------------------|
| 1 | Brand-new dApp, no inputs             | Sync finishes after ≤ 2 RPC calls; node ready in < 2 s.        |
| 2 | Inputs only **after** deploy          | Scans `[appDeploy … latest]`; RPC/time drop ≥ 80 %.            |
| 3 | Inputs **before** deploy              | Falls back to full scan; behaviour identical to current.       |                            |

This optimisation slashes first-run synchronisation time and RPC usage for new applications while preserving correctness for existing ones.

mpolitzer:
close #686

Skip scanning blocks for inputs of new applications when provably not necessary. This is done by confirming zero inputs on certain blocks with a call to getNumberOfInputs.

update: use the same strategy to sync outputs execution as well.

### DIFF
--- a/internal/evmreader/input.go
+++ b/internal/evmreader/input.go
@@ -7,11 +7,105 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/big"
 
 	. "github.com/cartesi/rollups-node/internal/model"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 )
+
+// Find the last block number that should be considered checked when scanning the blockchain
+// for inputs of this application. The next search should start from (lastBlockChecked + 1).
+//
+// The main purpose of this function is to reduce the range of blocks that need to be scanned,
+// avoiding unnecessary lookups. Currently, it is only used when handling new applications.
+//
+// Rules applied in order:
+// 1. Default fallback: the block before the InputBox was deployed.
+// 2. If the application has never received inputs, use the most recent block.
+// 3. If the application has no inputs since its deployment, use the deployment block.
+// 4. Otherwise, return the block before the InputBox deployment (conservative bound).
+//
+// Note: this function returns the *last block checked*, not the first block to search.
+// Example:
+//
+//	lastBlockChecked := findLastInputCheckBlock(...)
+//	nextSearchBlock  := lastBlockChecked + 1
+func findLastInputCheckBlock(app *appContracts, mostRecentBlockNumber uint64, mostRecentBlockNumberCallOpts *bind.CallOpts) uint64 {
+	var noiBig *big.Int
+	var err error
+	blockBeforeInputBox := app.application.IInputBoxBlock - 1
+
+	// find if the application has ever received any input. sync to present if not
+	noiBig, err = app.inputSource.GetNumberOfInputs(
+		mostRecentBlockNumberCallOpts,
+		app.application.IApplicationAddress,
+	)
+	if err != nil {
+		return blockBeforeInputBox
+	}
+	if noiBig.Uint64() == 0 {
+		return mostRecentBlockNumber
+	}
+
+	// find if the application has received an input since its deployment. sync to that block if not
+	// we'll need its deployment block number to do that
+	deploymentBlockNumberBig, err := app.applicationContract.GetDeploymentBlockNumber(mostRecentBlockNumberCallOpts)
+	if err != nil {
+		return blockBeforeInputBox
+	}
+
+	noiBig, err = app.inputSource.GetNumberOfInputs(&bind.CallOpts{
+		BlockNumber: deploymentBlockNumberBig,
+	},
+		app.application.IApplicationAddress,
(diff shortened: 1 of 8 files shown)


internal/evmreader/input.go
```diff
@@ -39,18 +85,27 @@ func (r *Service) checkForNewInputs(
 			"inputbox_address", inputBoxAddress,
 			"most recent block", mostRecentBlockNumber,
 		)
-		appsByLastInputCheckBlock := indexApps(byLastInputCheckBlock, inputBoxApps)
```
guidanoli, line 42: With this change, `byLastInputCheckBlock` would no longer be used.
Do you think it will be used in the future, and have decided to keep it?

internal/evmreader/output.go
```diff
@@ -21,11 +22,20 @@ func (r *Service) checkForOutputExecution(
...
+		BlockNumber: new(big.Int).SetUint64(mostRecentBlockNumber),
+	}
 
-		// Safeguard: Only check blocks starting from the block where the InputBox
-		// contract was deployed as Inputs can be added to that same block
-		lastOutputCheck := max(app.application.LastOutputCheckBlock, app.application.IInputBoxBlock)
+	for _, app := range apps {
+		lastOutputCheck := app.application.LastOutputCheckBlock
+		if lastOutputCheck == 0 { // New application. Find a safe start block to scan for outputs
+			lastOutputCheck = findSafeFirstInputBlockToScan(&app, mostRecentBlockNumber, mostRecentBlockNumberCallOpts) - 1
```
guidanoli: Why are we using `findSafeFirstInputBlockToScan` to scan for outputs?
Should we rename it to something more generic?

mpolitzer: > Why are we using findSafeFirstInputBlockToScan to scan for outputs?

We don't have a equivalent of `getNumberOfInputs` for Outputs.
Also this is a safe operation because:
```
firstInputBlockNumber <= firstOutputBlockNumber
```

> Should we rename it to something more generic?

The function does what its name implies. So I wouldn't change it.
What about a comment explaining why this is ok?

internal/evmreader/input.go
mpolitzer: Nice catch.
Lets remove it and add it back if we ever need it in the future.

internal/evmreader/output.go
guidanoli: This is only true if only valid output roots are finalized.

guidanoli: Here are some counterexamples:

- The authority owner's PK got stolen and approved an invalid outputs root
- A majority of the quorum submitted an invalid outputs root
- Only an invalid commitment was submitted to PRT and won by timeout

Since v2 allows for empty epochs, an application could have an invalid outputs root finalized but no inputs received; so it would be possible for an `OutputExecuted` event to be emitted before an `InputAdded` event for that app.

mpolitzer: Interesting. So the assumption is only valid for a working application.
But may not hold for a broken one.

I am thinking if the node should care about those cases.
Because all bets are off if the optimistic-rollup doesn't have a honest validator.

My gut feeling is that it shouldn't. The application is broken beyond repair.
But we should still detect something is wrong and stop processing this application.

@vfusco have you thought about this case?

guidanoli: I agree. If an output is executed before an input is added, the app is broken.
I understand if the node treats this as a "doesn't care" state.

vfusco (approved):

PR #688, opened 2025-07-21
Refactor machine pkg

vfusco:


pkg/emulator/machine.go
```diff
@@ -313,30 +313,38 @@ func (m *Machine) WriteReg(r RegID, value uint64) error {
...
+		data = make([]byte, cLen)
 		err = newError(C.cm_receive_cmio_request(
 			m.ptr,
 			&cCmd,
 			&cReason,
-			(*C.uint8_t)(unsafe.Pointer(&buf[0])),
+			(*C.uint8_t)(unsafe.Pointer(&data[0])),
 			&cLen,
 		))
 	})
```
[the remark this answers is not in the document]
vfusco: This is how the emulator code does it on it's own code. https://github.com/cartesi/machine-emulator/blob/v0.19.0/src/clua-i-virtual-machine.cpp#L840

pkg/machine/machine.go
```diff
@@ -0,0 +1,194 @@
...
+	ErrMachineInternal            = errors.New("machine internal error")
+	ErrDeadlineExceeded           = errors.New("machine operation deadline exceeded")
+	ErrCanceled                   = errors.New("machine operation canceled")
+	ErrOrphanServer               = errors.New("machine server was left orphan")
+	ErrNotAtManualYield           = errors.New("not at manual yield")
+	ErrException                  = errors.New("last request yielded an exception")
+	ErrRejected                   = errors.New("last request yielded as rejected")
+	ErrHalted                     = errors.New("machine halted")
+	ErrOutputsLimitExceeded       = errors.New("outputs limit exceeded")
+	ErrReachedTargetMcycle        = errors.New("machine reached target mcycle")
```
mpolitzer, line 41: Should reaching the target mcycle be an error?
It will be a regular occurrence when calculating the computation hash on PRT

mpolitzer (approved):

pkg/machine/machine.go
vfusco: Yeah the name is not good as well. We can revisit this when we get there.

PR #692, opened 2025-08-12
feat(cli): Check Inputbox contract existence on application register

### LINKED ISSUE #530: Check Inputbox contract existence on application register
## 📚 Context

~~EVM Reader reads inputs from a InputBox contract. But it do not verify if the contract is actually deployed or it is at the specified block. It should check and fail if the contract is absent or the informed `inputBoxDeplymentBlockNumber` is actually different from the actual one~~

We need to check if the application being registered was a properly deployed, pointing to the correct inputbox address.

## ✔️ Solution

~~Add a check at Evm Reader Service start~~

use `eth_getCode` to check the inputbox contract existence

mpolitzer:
close #530

### DIFF
--- a/cmd/cartesi-rollups-cli/root/app/register/register.go
+++ b/cmd/cartesi-rollups-cli/root/app/register/register.go
@@ -206,6 +206,17 @@ func run(cmd *cobra.Command, args []string) {
 		inputBoxBlockNumber = block.Uint64()
 	}
 
+	// ensure there is a contract deployed at the input box address
+	hasCode, err := hasCodeAt(ctx, *inputBoxAddress)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to probe input box address for contract: %v\n", err)
+		os.Exit(1)
+	}
+	if !hasCode {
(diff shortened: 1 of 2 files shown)


cmd/cartesi-rollups-cli/root/app/register/register.go
```diff
@@ -229,6 +240,25 @@ func getTemplateHash(
 	return ethutil.GetTemplateHash(ctx, client, appAddress)
 }
 
+func hasCodeAt(
```
vfusco, line 276: Ethutil?

vfusco (changes requested):

cmd/cartesi-rollups-cli/root/app/register/register.go
mpolitzer: I disagree. Most of the code base passes a constructed `client` around.
This file is the different one that creates a `client` for each blockchain interaction.
I think we should leave the PR as is and later, when we have time, refactor this code to do the same.
Pass `client` around and delete this function later (becomes a direct call to `client.CodeAt`).

vfusco (approved):

PR #705, opened 2025-10-09
feat(merkle): add builder on merkle trees data structure

mpolitzer:
merkle builder data structure re-implemented in Go, from Dave repository.


vfusco:
Merged into feature/prt-draft-reader-mode

PR #751, opened 2026-03-09
Add initial integration tests 

vfusco:


.github/workflows/clean-up-images.yml
```diff
@@ -9,22 +9,23 @@ on:
...
         with:
           organization: cartesi
           container: ${{ matrix.image }}
           token: ${{ secrets.GITHUB_TOKEN }}
-          prune-untagged: false
+          prune-untagged: true
           keep-last: 0
           prune-tags-regexes: |
             ^pr-${{ github.event.number }}$
+            ^ci-
```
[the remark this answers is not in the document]
vfusco: fixed.

test/integration/cli_helpers_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	var app struct {
+		IApplicationAddress string `json:"iapplication_address"`
+		IConsensusAddress   string `json:"iconsensus_address"`
+	}
+	if err := json.Unmarshal([]byte(out), &app); err != nil {
+		return "", fmt.Errorf("parse deploy output: %w", err)
+	}
+	return app.IApplicationAddress, nil
+}
```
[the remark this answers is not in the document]
vfusco: fixed.

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,55 @@
...
+export PATH="/opt/go/bin:/build/cartesi/go/rollups-node:$PATH"
+
+# Smoke-check: verify the CLI binary is on PATH before running the suite.
+which cartesi-rollups-cli || { echo "ERROR: cartesi-rollups-cli not found on PATH"; exit 1; }
+
+# Timeout must be less than the CI job timeout-minutes (60) to produce
+# a useful go test panic instead of an abrupt CI kill.
+go test -count=1 -v -timeout 55m \
+  -ldflags "-r /opt/cartesi/lib" \
+  --tags=endtoendtests ./test/integration/...
```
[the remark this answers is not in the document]
vfusco: fixed.

pkg/machine/implementation.go
```diff
@@ -245,6 +258,23 @@ func (m *machineImpl) Close() error {
...
+// waitForExit polls until the process with the given PID has exited or the
+// timeout elapses. Returns true if the process exited within the timeout.
+func waitForExit(pid uint32, timeout time.Duration) bool {
+	deadline := time.Now().Add(timeout)
+	for time.Now().Before(deadline) {
+		proc, err := os.FindProcess(int(pid))
+		if err != nil {
+			return true
+		}
+		if proc.Signal(syscall.Signal(0)) != nil {
```
renatomaia, line 271:
```suggestion
		// try send no-op signal 0 to check process is still receiving signals.
		if proc.Signal(syscall.Signal(0)) != nil {
```

test/integration/anvil_helpers_test.go
```diff
@@ -0,0 +1,291 @@
+// (c) Cartesi and individual authors (see AUTHORS)
```
mpolitzer, line 1: We are replicating some functionality from: `pkg/ethutil/anvil.go` here.
Not necessarily now, but we should unify them both.

renatomaia (dismissed):

pkg/machine/implementation.go
vfusco: fixed

renatomaia (dismissed):

test/integration/polling_helpers_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/cartesi/rollups-node/internal/model"
+)
+
+// pollUntil polls fn at the given interval until it returns true or the context
+// is cancelled. Returns an error if the context times out.
+func pollUntil(ctx context.Context, interval time.Duration, fn func() (bool, error)) error {
```
mpolitzer, line 19: This is a nitpick, but I would call this `Retry`.

Intuitively, my first expectation for a `time.Duration` for `pollUntil` would be some sort of timeout, not the interval.

Also, something to consider is having a back-off mechanism or maybe using a lib for this such as: [github.com/codeGROOVE-dev/retry](https://pkg.go.dev/github.com/codeGROOVE-dev/retry), `retry.DoWithData` in particular.

test/integration/lifecycle_test.go
```diff
@@ -0,0 +1,411 @@
...
+// Used to fund application contracts so vouchers can transfer ETH during tests.
+const oneEtherWei = "0xde0b6b3a7640000"
+
+// Expected output/report counts per echo-dapp input.
+// The ioctl-echo-loop dapp produces 3 outputs (voucher + delegatecall voucher + notice)
+// and 1 report per accepted input. These constants keep assertions in sync with dapp behavior.
+const (
+	echoOutputsPerInput           = 3 // Voucher + DelegateCallVoucher + Notice
+	echoReportsPerInput           = 1
+	rejectOutputsPerAcceptedInput = 2 // Voucher + Notice (no DelegateCallVoucher)
```
mpolitzer, line 41: This doesn't look correct to me. There should be either 3 (accept) or 0 (reject) outputs, nothing in between. Or are those results for different applications (not default echo-dapp?)

 More on `runRejectExceptionLifecycleTest`

test/integration/lifecycle_test.go
```diff
@@ -0,0 +1,411 @@
...
+	}()
+
+	// --- Verify off-chain results: only accepted inputs produce outputs ---
+
+	t.Logf("Checking outputs — only accepted inputs should produce them (no outputs from %s input #1)...",
+		cfg.FailStatus)
+	outputsResp, err := readOutputs(ctx, cfg.AppName)
+	require.NoError(err, "read outputs")
+	numAccepted := uint64(2)
+	require.Equal(numAccepted*rejectOutputsPerAcceptedInput, outputsResp.Pagination.TotalCount,
```
mpolitzer, line 257: I instrumented the test and got:
```
numAccepted=2 rejectOutputsPerAcceptedInput=2
```

But shouldn't the correct answer be:
```
numAccepted=2 rejectOutputsPerAcceptedInput=3
```
instead?

internal/jsonrpc/api/decode.go
```diff
@@ -0,0 +1,276 @@
...
+	if len(input.RawData) < 4 {
+		return &DecodedInput{Input: input}, fmt.Errorf("error: input needs at least 4 bytes")
+	}
+
+	method, ok := parsedAbi.Methods["EvmAdvance"]
+	if !ok {
+		return &DecodedInput{Input: input}, fmt.Errorf("EvmAdvance method not found in ABI")
+	}
+
+	err := method.Inputs.UnpackIntoMap(decoded, input.RawData[4:])
```
mpolitzer, line 69: should we check that function selector matches as well?

mpolitzer (dismissed):
This is great! Having robust integration tests is something we needed for a while.

test/integration/anvil_helpers_test.go
vfusco: Ok. It's on my list to review package ethutil. I'll revisit this there. Thank you

test/integration/polling_helpers_test.go
vfusco: I think `pollUntil` works well here because it clearly indicates we're polling at intervals until a condition is met. The `interval` parameter is indeed the polling frequency (not a timeout), with the context handling the overall timeout. `Retry` might imply retrying on failure, but this is more about repeated checking until success, so the polling terminology feels more accurate to me.

Thanks for the link. I'll take a look.

test/integration/lifecycle_test.go
vfusco: Thanks for double-checking! The constant is actually correct at 2. The reject-loop-dapp is built with --vouchers=1 --notices=1 and no --delegate-call-vouchers flag (defaults to 0), so each accepted input produces 1 voucher + 1 notice = 2 outputs. This is intentional — rejectOutputsPerAcceptedInput is only used in
  the reject/exception lifecycle tests, separate from the echo-dapp which produces 3 (adds a DelegateCallVoucher). Having different output counts between the two test paths gives us better coverage. Your instrumented result of rejectOutputsPerAcceptedInput=2 confirms it matches.

test/integration/lifecycle_test.go
vfusco: These constants are not for the default echo-dapp — they're specific to the reject-loop-dapp, which is a separate application built with different output flags (--vouchers=1 --notices=1, no --delegate-call-vouchers). The echo-dapp uses echoOutputsPerInput = 3 (line 39). The reject-loop intentionally produces a  different count (2) so we get broader test coverage across the two test paths.

The ioctl-echo-loop --reject=1 flag tells the DApp to reject the input at index 1 — it still emits outputs/reports before rejecting, but the machine discards them upon rejection. The reject lifecycle test sends 3 inputs (indices 0, 1, 2), expects index 1 to be rejected, and verifies that only the 2 accepted inputs (0 and 2) produce outputs — so numAccepted=2 * rejectOutputsPerAcceptedInput=2 = 4 total outputs.

internal/jsonrpc/api/decode.go
vfusco: fixed

test/integration/lifecycle_test.go
mpolitzer: ok, makes sense. It seems I got tricked by the comment then.

https://github.com/cartesi/rollups-node/pull/751/changes#diff-66a04d3be01bea8feeec73fc5e8a7d911ecbdad47c098e786d96c34bfe1996e4R35-R37

mpolitzer (approved):

PR #755, opened 2026-03-12
Fix evmreader reliability: subscription retry/liveness, idempotent inputs, input count validation, and unified fetch logic

vfusco:


internal/evmreader/evmreader.go
```diff
@@ -109,107 +140,184 @@ func (r *Service) setApplicationInoperable(ctx context.Context, app *Application
...
+			}
+		}
 
+		// Build Contracts (adapters are cached per application address)
+		var apps []appContracts
+		var daveConsensusApps []appContracts
+		var iconsensusApps []appContracts
+		for _, app := range runningApps {
+			addr := app.IApplicationAddress
+			cached, ok := adapterCache[addr]
```
mpolitzer: I would rename this `ok` to `hit`, `isCacheValid` or something of the sort.

The way it is, one would expect the `ok` to only be related only to the map access. But it isn't. So the `ok = false` from line 253 is a bit surprising. And also wounder why `if !ok {...` is not an else.

internal/evmreader/input_indexing_test.go
```diff
@@ -0,0 +1,427 @@
...
+	"math"
+	"testing"
+	"testing/quick"
+
+	. "github.com/cartesi/rollups-node/internal/model"
+	"github.com/stretchr/testify/require"
+)
+
+// makeInput creates a test input with the given index and block number.
+func makeInput(index, blockNumber uint64) *Input {
```
mpolitzer, line 16: Something to improve across the codebase in an opportune is to use the builders.go for the tests such as this file. (inputs and epochs in particular)

NOTE: doesn't have to be on this PR

mpolitzer (dismissed):

internal/evmreader/input_indexing_test.go
mpolitzer: Actually, there is already an issue for this: https://github.com/cartesi/rollups-node/issues/756

internal/evmreader/evmreader.go
vfusco: fixed

mpolitzer (dismissed):

mpolitzer (dismissed):

renatomaia (dismissed):

renatomaia (approved):

mpolitzer (approved):

PR #757, opened 2026-03-13
Feature/narrow claimer event scanning

mpolitzer:
Draft of claimer update to use event counters.

We'll rework the algorithm instead, so don't merge this for now.


mpolitzer:
Ideas diverged too much from next/2.0. Maybe revisit on the future.

PR #761, opened 2026-03-16
fix(inspect): limit request body size

vfusco:


internal/inspect/inspect.go
```diff
@@ -119,12 +125,24 @@ func (inspect *Inspector) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	dapp = r.PathValue("dapp")
 	if r.Method == "POST" {
-		payload, err = io.ReadAll(r.Body)
+		// Limit the request body to the machine's CMIO RX buffer size.
+		// Payloads larger than this are rejected by the machine, so reading
+		// them into memory would only waste resources. We read maxPayloadSize+1
+		// bytes so we can distinguish "exactly at limit" from "over limit".
+		limitedReader := io.LimitReader(r.Body, maxPayloadSize+1)
```
mpolitzer, line 132: Is `maxPayloadSize` in raw bytes or hex encoded? machine can fit `2x` if hex encoded.

internal/inspect/inspect.go
```diff
@@ -21,6 +21,12 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
+// maxPayloadSize is the maximum allowed inspect request body size.
+// This matches the Cartesi Machine's CMIO RX buffer (2 MiB, defined as log2 size 21
+// in the machine emulator's pma-defines.h). Payloads larger than this are rejected
+// by the machine anyway, so there is no reason to read them into memory.
+const maxPayloadSize = 1 << 21 // 2 MiB
```
mpolitzer, line 28: I find it easier to read it like this, but its a nitpick:
```suggestion
const maxPayloadSize = 2 << 20 // 2 MiB
```
```
x << 10 (x KiB) 
x << 20 (x MiB)
x << 30 (x GiB)
```

renatomaia (approved):

internal/inspect/inspect.go
vfusco: I see your point. But I'll leave like it's on the emulator.

```
 PMA_CMIO_RX_BUFFER_LOG2_SIZE_DEF 21
 1 << emulator.CmioRxBufferLog2Size
 ```

internal/inspect/inspect.go
vfusco: It's raw bytes end-to-end, no hex encoding. The HTTP body (`[]byte`) flows unchanged through the manager, machine, and emulator all the way to `cm_send_cmio_response `in the C API. The CLI also sends `application/octet-stream` (its `--hex` flag just decodes hex to bytes before sending). So the 2 MiB limit (`1 << 21`) matches the machine's CMIO RX buffer exactly.

mpolitzer (approved):

PR #770, opened 2026-03-25
Proper fix for the count() over() problem

vfusco:


internal/repository/postgres/application.go
```diff
@@ -511,6 +511,39 @@ func (r *PostgresRepository) ListApplications(
...
+
+	countStmt := table.Application.SELECT(postgres.COUNT(postgres.STAR)).FROM(fromClause)
+	if len(conditions) > 0 {
+		countStmt = countStmt.WHERE(postgres.AND(conditions...))
+	}
+	total, err := countFromTx(ctx, tx, countStmt)
+	if err != nil {
+		return nil, 0, err
+	}
+
```
renatomaia, line 549: Is it worth to return an empty result if `total == 0` just to avoid the following query in the DB?

renatomaia (dismissed):

internal/repository/postgres/application.go
vfusco: Good call. Added the early return on `total == 0` across all 10 list functions. It avoids the second query (with JOINs, ORDER BY, pagination) when there are no matching rows, which is especially common with filtered queries.

renatomaia (dismissed):

mpolitzer (approved):

PR #775, opened 2026-04-30
feat: use RunAndCollectRootHashesNew

mpolitzer:
Reworked `RunAndCollectRootHashes` to have a fixed-point hash when collecting.
This mirrors the emulator function `cm_collect_mcycle_root_hashes`.

Additionally we use `prevMachineHash` (the checkpoint) on `REJECT`ed inputs instead of what was collected.

The logic in pseudo code becomes:
```
hashes := collect_mcycle_root_hashes()
if accept {
  // use hashes as is.
} else { // reject
  revert()
  hashes[#hashes-1] = REVERT_ROOT_HASH
}
```

Later, for state hashes:
  - Take each hash from the input with a repetition of 1, except the last one,
  -  which takes whatever is left from the total.


mpolitzer:
will be implemented here: https://github.com/cartesi/rollups-node/pull/791

PR #778, opened 2026-05-20
validator hardening

mpolitzer:
`drop 6aef485900512cf257c6905a821d0aadc07eff20 "fix(validator): guard against using outputs from across epochs"` - handled by bump contracts v3.

### DIFF
--- a/internal/validator/validator.go
+++ b/internal/validator/validator.go
@@ -470,8 +470,15 @@ func (s *Service) computeMerkleTreeAndProofs(
 
 	// if there are no outputs
 	if len(epochOutputs) == 0 {
+		// and there should be a previous epoch (guard against DB corruption)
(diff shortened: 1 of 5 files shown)


internal/validator/validator.go
```diff
@@ -210,7 +220,7 @@ func (s *Service) validateApplication(ctx context.Context, app *Application) err
 		}
 
 		// DaveConsensus can have empty epochs. Authority and Quorum don't.
```
renatomaia: Should this comment move to the else below?

renatomaia (dismissed):
Should the commit message be `fix(validator): check for context cancellation before processing *EACH* epoch`?

vfusco (approved):

renatomaia (approved):

PR #779, opened 2026-05-22
bump contracts 3.0.0 alpha

vfusco:


pkg/ethutil/withdrawal_config.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	return nil
+}
+
+// withdrawalConfigArgs is sourced from the abigen-generated
+// IApplicationFactory ABI so the encoding stays in lockstep with the
+// contract definition. The tuple type is taken from
+// `calculateApplicationAddress` (unique signature, no overload ambiguity).
+var withdrawalConfigArgs = func() abi.Arguments {
```
mpolitzer, line 62: I've never seen a panic in this context.

This is a theoretical question since we know the ABI is there in the source code, but it got me curious as to what would happen if this got triggered.

Would the panic happen during startup?

mpolitzer (dismissed):

renatomaia (dismissed):

mpolitzer (dismissed):

guidanoli (dismissed):
I asked Claude Opus 4.8 to summarize the blockchain-touching parts of the claimer and the assumptions it has about the contracts, and they all seem compatible with the version 3.0.0-alpha.6 of the Rollups contracts.

guidanoli (approved):

mpolitzer (approved):

PR #793, opened 2026-07-30
Feature/json rpc api improvements

renatomaia:
# JSON-RPC API Improvements

This PR implements nine improvements identified as F1 to F9 as described [here](https://github.com/user-attachments/files/30548342/jsonrpc-api-improvements-task.html), and other improvements described in reviews [here](https://github.com/user-attachments/files/31092856/jsonrpc-api-improvements-pr793-review.html) and [here](https://github.com/user-attachments/files/31234724/jsonrpc-api-improvements-pr793-merge-gate.html).

## Divergences from Proposed Fixes

1. H1 requires empty batches and batches over 100 entries to return error code -32600. The implementation instead defines and returns -32040 as suggested [here](https://json-rpc.dev/docs/reference/error-codes) to indicate _invalid batch request_.
2. Y5 in the review requested true one-round-trip implementation, however the proposed fix was not possible because the repository queries still return `COUNT(*)` rather than `ErrNotFound` for an unknown application.
3. Y7 in the review explicitly requested moving single-request path logging the client-controlled method name from `Info` level to `Debug` level, but the original behavior was kept because one `Info` line was allowed per batch in the description of feature F1.
4. F2 and F6 describe the use of error codes -32001 for missing resources and -32002 for unknown applications from the original implementation. But such codes were replaced by -31001 and -31002 to avoid reserved server error codes as suggested [here](https://json-rpc.dev/docs/reference/error-codes).

## Commit relation to Features and Improvements

The table below lists the commits, and column `ID` indicates which improvement is introduced.

| ID   | Commit   | Change   | Component  | Description |
| ---- |:--------:|:--------:| ---------- | ----------- |
| F1   | 733a8f4f | feat     | jsonrpc    | add support for JSON-RPC 2.0 batch requests |
| H6   | e954e1a9 | feat     | jsonrpc    | impose the same response size budget of batch requests to single requests |
|      | af297384 | refactor | jsonrpc    | remove unused field 'data' from error responses |
| F2   | da49cc46 | feat     | jsonrpc    | add operation to get epoch by a virtual contiguous index |
| F3   | d9bd6989 | feat     | jsonrpc    | add operation to get Node info like its chain ID, version, and default block |
| F4   | dca72551 | feat     | jsonrpc    | add inclusive index ranges to list epochs, inputs, outputs, and reports |
| F5   | 28ecb425 | feat     | jsonrpc    | allows to filter output by execution and multiple selectors |
| H2   | 20375007 | perf     | repository | add DB index to improve filter output by execution |
|      | 8f63de24 | feat     | cli        | allows to filter output by execution and multiple selectors |
| F6   | 77448eaf | feat     | jsonrpc    | add methods to get the count of executed and pending outputs |
| F7   | f294cdc9 | feat     | jsonrpc    | support listing epochs with multiple statuses |
|      | c12dd7b6 | feat     | cli        | support listing epochs with multiple statuses |
| F8   | 94d954be | fix      | jsonrpc    | report 256-bit integer fields in OpenRPC specification |
| F9   | 4906bcd3 | refactor | jsonrpc    | rename cartesi_getMatchAdvanced as cartesi_getMatchAdvance |
|      | 31568e8e | test     | jsonrpc    | add tests for positional decoding of parameters |
| Y1   | 3a769b3c | fix      | repository | avoid invalid SQL when listing outputs with empty type list |
| Y6   | 9ba37eaf | test     | jsonrpc    | add tests for some use cases of listing operations |
| Y7#5 | bb9906e6 | perf     | jsonrpc    | parse JSON-RPC API spec on service initialization |
| Y7#7 | 2911c1b8 | docs     | repository | add comment to clarify expected behavior of module API |
| H1   | 2451cedf | docs     | jsonrpc    | add warnings and recommendations on how to use the API |
| H4   | 7531899d | docs     | jsonrpc    | improve documentation on response size limit |
|      | 65f2d404 | fix      | jsonrpc    | skip ignored fields on positional decoding of parameters |
| Y6   | ebb3d958 | test     | jsonrpc    | add more tests on parsing parameters |
| Y7#8 | 9bcb40d6 | style    | cli        | avoid line length violation |
| Y7#9 | d79b0758 | test     | jsonrpc    | add test to enforce log of method call |
| H6   | 54231eea | feat     | jsonrpc    | impose the same listed items limit for all requests in a batch |
| N1   | ac78085e | feat     | jsonrpc    | add a dispatch timeout to prevent reaching the write timeout to avoid incomplete responses |
| N2   | f8c17233 | fix      | jsonrpc    | refuse offsets that are too large to be handled correctly |
| N3   | 96c0a146 | fix      | jsonrpc    | avoid logging canceled context errors due to disconnected clients |
| N4   | 7c6012dc | feat     | jsonrpc    | truncate long method names to avoid flooding the log |
| N5   | 01aeb98a | feat     | jsonrpc    | report panic errors of individual methods in a request batch |
| N6   | 4e9f329d | feat     | jsonrpc    | avoid overlogging decoding errors of DApps inputs/outputs |
| N7   | 17f15423 | fix      | jsonrpc    | avoid response with different ID value due to decoding errors |
| N8   | 726d8444 | docs     | jsonrpc    | document the potential costs of DB transversal due to large offsets |
| R1   | d4ea032e | test     | jsonrpc    | add endpoint-level tests for 'cartesi_listOutputs' |
| R3   | b79a9988 | perf     | jsonrpc    | add DB index to improve filter of executed outputs |
| R4   | 55a3eb0f | docs     | jsonrpc    | document wort-case memory scenario for responses |
| R8   | 2cdb69bd | test     | jsonrpc    | avoid different test package setup to interfere with one another |
|      | 72003b4a | style    | jsonrpc    | avoid lint errors |
| R2   | a89cd031 | style    | cli        | avoid lint errors |
|      | 4308df98 | fix      | jsonrpc    | report JSON contents even on incomplete/malformed responses |
|      | 57f44491 | fix      | jsonrpc    | avoid corrupting the offset maximum reported by 'rpc.discover' |
|      | 7558af64 | fix      | jsonrpc    | improve error messages reported to the client |


renatomaia:
I added two commits addressing the first and third issues of the Copilot review above.

I’m deferring the second issue for now. The 10 MiB response budget is intended as a practical memory bound rather than a byte-exact HTTP body limit. Batch delimiters and replacement error responses may add a small, bounded overhead, which we consider negligible for the current purpose.

I’m also deferring the fourth issue. We have decided not to introduce a new database migration at this time, based on our current deployment and usage patterns. Fresh databases will include the indexes, but existing schema-version-1 databases will not receive them automatically. We can add a versioned migration later if supporting those upgrades becomes necessary.

internal/jsonrpc/jsonrpc.go
```diff
@@ -83,6 +96,54 @@ var jsonrpcHandlers = dispatchTable{
...
+
+// writeRPCError sends a generic error response for internal errors.
+func (s *Service) writeRPCError(w http.ResponseWriter, id any, code int, message string) bool {
+	err := writeRPCError(w, id, code, message, nil)
+	return s.handleWriteResponse(err)
+}
+
+func (s *Service) dispatchOneRequest(w io.Writer, r *http.Request, req RPCRequest) error {
+	switch req.ID.(type) {
+	case nil, string, float64:
```
mpolitzer: What about integer ID types? Are they invalid?

internal/jsonrpc/jsonrpc.go
```diff
@@ -83,6 +96,54 @@ var jsonrpcHandlers = dispatchTable{
...
+	default:
+		return writeRPCError(w, nil, JSONRPC_INVALID_REQUEST, "invalid request", nil)
```
mpolitzer: Should we qualify the error?

```
-return writeRPCError(w, nil, JSONRPC_INVALID_REQUEST, "invalid request", nil)
+return writeRPCError(w, nil, JSONRPC_INVALID_REQUEST, "invalid request ID", nil)
```

internal/jsonrpc/jsonrpc.go
```diff
@@ -83,6 +96,54 @@ var jsonrpcHandlers = dispatchTable{
...
+	}
+	if req.JSONRPC != "2.0" || req.Method == "" {
+		return writeRPCError(w, req.ID, JSONRPC_INVALID_REQUEST, "invalid request", nil)
+	}
```
mpolitzer, line 228: You can check RPC version here and leave the method check to the if below.
```
	if req.JSONRPC != "2.0" {
		return writeRPCError(w, req.ID, JSONRPC_INVALID_REQUEST, "unsupported JSON-RPC version != 2.0", nil)
	}
```

.

internal/jsonrpc/jsonrpc.go
```diff
@@ -83,6 +96,54 @@ var jsonrpcHandlers = dispatchTable{
...
+
+	result, err := fn(s, r, req)
+	if err == nil {
+		return writeRPCResult(w, req.ID, result)
+	}
+
+	var rpcErr *RPCError
+	if errors.As(err, &rpcErr) {
+		return writeRPCError(w, req.ID, rpcErr.Code, rpcErr.Message, rpcErr.Data)
+	}
```
mpolitzer, line 251: My understanding is that everything from here on below will only trigger if the error returned above is **not** convertible to a `rpcErr`. It has to be some other kind.

Is this intended? Should the log be lifted to between the `err == nil` and the `error.As` checks?

internal/jsonrpc/jsonrpc.go
```diff
@@ -97,17 +158,99 @@ func (s *Service) handleRPC(w http.ResponseWriter, r *http.Request) {
...
+
+	case '[':
+		w.Header().Set("Content-Type", "application/json")
+		var reqSeq []json.RawMessage
+		if err := json.Unmarshal(body, &reqSeq); err != nil {
+			s.writeRPCError(w, nil, JSONRPC_PARSE_ERROR, "invalid request batch")
+			return
+		}
+		if len(reqSeq) == 0 || len(reqSeq) > MAX_BATCH_SIZE {
+			s.writeRPCError(w, nil, JSONRPC_INVALID_BATCH, "invalid request batch size (expected [1..100])")
```
mpolitzer: **untested**
```
s.writeRPCError(w, nil, JSONRPC_INVALID_BATCH, fmt.Sprintf("invalid request batch size (expected [1..%v])", MAX_BATCH_SIZE))

```

internal/jsonrpc/jsonrpc.go
```diff
@@ -97,17 +158,99 @@ func (s *Service) handleRPC(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		s.Logger.Info("Dispatching RPC request", "method", req.Method)
+		err := s.dispatchOneRequest(w, r, req)
+		s.handleWriteResponse(err)
+
+	case '[':
+		w.Header().Set("Content-Type", "application/json")
+		var reqSeq []json.RawMessage
```
mpolitzer, line 354: Any reason why to not decode directly to `[]RPCRequest` here? Is it so that each individual request has a chance to succeed?

maybe worth a comment.

internal/jsonrpc/jsonrpc.go
```diff
@@ -97,17 +158,99 @@ func (s *Service) handleRPC(w http.ResponseWriter, r *http.Request) {
...
+			if i > 0 && !s.writeByte(w, ',') {
+				return
+			}
+
+			var responded bool
+			var req RPCRequest
+
+			switch r.Context().Err() {
+			case context.Canceled:
+				return
```
mpolitzer, line 384: Does closing the array and `,` logic work correctly with this early return?

Maybe a `defer s.writeByte(w, ']')` after opening would make this clearer.

internal/jsonrpc/jsonrpc.go
```diff
@@ -97,17 +158,99 @@ func (s *Service) handleRPC(w http.ResponseWriter, r *http.Request) {
...
+			}
+		}
+		s.writeByte(w, ']')
+
+	default:
+		w.Header().Set("Content-Type", "application/json")
+		if json.Valid(body) {
+			s.writeRPCError(w, nil, JSONRPC_INVALID_REQUEST, "invalid request")
+		} else {
+			s.writeRPCError(w, nil, JSONRPC_PARSE_ERROR, "Parse error")
```
mpolitzer, line 413: I know the codebase is mixed on this, but some messages are Capitalized like this one, and some are not like the other branch (`if`).

Lint suggest preferring lower case: `error strings should not be capitalized or end with punctuation, question mark, or exclamation mark`

mpolitzer (commented):
Found a couple nits, will finish the review monday

vfusco (dismissed):

mpolitzer:
I added a few comments to 97c92532700fcdac245827d8d5a879677b963c90, they didn't show up here on the Conversation tab. I'll see if I can redo them.

internal/jsonrpc/jsonrpc.go
```diff
@@ -25,65 +31,285 @@ var discoverSpec embed.FS
...
+	}
+	s.Logger.Error(message, "err", err)
+	return newRPCError(JSONRPC_INTERNAL_ERROR, "Internal server error")
+}
+
+func (s *Service) handleRequest(w io.Writer, r *http.Request, req RPCRequest) error {
+	if !validRPCID(req.ID) {
+		return writeRPCError(w, nil, JSONRPC_INVALID_REQUEST, "invalid request")
+	}
+	if req.JSONRPC != "2.0" || req.Method == "" {
```
mpolitzer: Empty and invalid method is already handled below, on the `!ok` branch.

```suggestion
	if req.JSONRPC != "2.0" {
```

internal/jsonrpc/jsonrpc.go
```diff
@@ -25,65 +31,285 @@ var discoverSpec embed.FS
...
+	if errors.Is(err, context.Canceled) {
+		return err
+	}
+	if errors.Is(err, context.DeadlineExceeded) && errors.Is(r.Context().Err(), context.DeadlineExceeded) {
+		s.Logger.Warn("RPC method dispatch timeout", "method", truncatedMethod(req.Method))
+		return writeRPCError(w, req.ID, JSONRPC_TIMEOUT_ERROR, "Request timed out")
+	}
+
+	var rpcErr *RPCError
+	if errors.As(err, &rpcErr) {
```
mpolitzer, line 247: My understanding is that everything from here on below will only trigger if the error returned above is not convertible to a rpcErr. It has to be some other kind.

Is this intended? Should the log be lifted to between the err == nil and the error.As checks?

internal/jsonrpc/jsonrpc.go
```diff
@@ -97,17 +323,89 @@ func (s *Service) handleRPC(w http.ResponseWriter, r *http.Request) {
...
+		var req RPCRequest
+		if err := json.Unmarshal(body, &req); err != nil {
+			s.writeRPCError(w, nil, JSONRPC_PARSE_ERROR, "invalid request")
+			return
+		}
+		s.Logger.Info("Dispatching RPC request", "method", truncatedMethod(req.Method))
+		s.dispatchOneRequest(w, r, req, budgetResp)
+
+	case '[':
+		var reqSeq []json.RawMessage
```
mpolitzer, line 354: Any reason why to not decode directly to []RPCRequest here? Is it so that each individual request has a chance to succeed?

maybe worth a comment.

internal/jsonrpc/jsonrpc.go
```diff
@@ -97,17 +323,89 @@ func (s *Service) handleRPC(w http.ResponseWriter, r *http.Request) {
...
+		if err := json.Unmarshal(body, &reqSeq); err != nil {
+			s.writeRPCError(w, nil, JSONRPC_PARSE_ERROR, "invalid request batch")
+			return
+		}
+		if len(reqSeq) == 0 || len(reqSeq) > MAX_BATCH_SIZE {
+			s.writeRPCError(w, nil, JSONRPC_INVALID_BATCH, "invalid request batch size (expected [1..100])")
```
mpolitzer: s.writeRPCError(w, nil, JSONRPC_INVALID_BATCH, fmt.Sprintf("invalid request batch size (expected [1..%v])", MAX_BATCH_SIZE))

internal/jsonrpc/jsonrpc.go
```diff
@@ -97,17 +323,89 @@ func (s *Service) handleRPC(w http.ResponseWriter, r *http.Request) {
...
+			if i > 0 && !s.writeByte(w, ',') {
+				return
+			}
+
+			var responded bool
+			var req RPCRequest
+
+			switch r.Context().Err() {
+			case context.Canceled:
+				return
```
mpolitzer, line 384: Does closing the array and , logic work correctly with this early return?

Maybe a defer s.writeByte(w, ']') after opening would make this clearer.

internal/jsonrpc/jsonrpc.go
renatomaia: Fixed.

internal/jsonrpc/jsonrpc.go
renatomaia: Fixed.

internal/jsonrpc/jsonrpc.go
renatomaia: Added a comment describing the reason for this.

internal/jsonrpc/jsonrpc.go
renatomaia: Fixed.

internal/jsonrpc/jsonrpc.go
renatomaia: Added a comment.

internal/jsonrpc/jsonrpc.go
renatomaia: The context is canceled when the client connection is closed or other cases where we cannot provide a response (see [here](https://pkg.go.dev/net/http#Request.Context)).

internal/jsonrpc/jsonrpc.go
renatomaia: Fixed.

mpolitzer (approved):
This is great work!

vfusco (approved):

PR #794, opened 2026-08-14
Feature deterministic input outcomes and replay verification

vfusco:
This PR makes the node's persisted history deterministic — the foundation for the dispute-capable PRT node. The machine layer now returns only the four guest-decided outcomes (accepted, rejected, exception, halted) as results; everything else — cycle ceilings, deadlines, payload limits, emulator faults — travels as an error and is never persisted, so two nodes can no longer durably disagree about the same input. The database enforces the same rule with a narrowed status enum, atomic result writes, and immutability constraints, and a new replay engine re-executes stored history at startup, fencing any application whose records contradict actual execution while healthy applications continue. Machine state storage and recovery (snapshots, retention, state serving) is deliberately out of scope and follows in a separate changeset.


internal/repository/postgres/schema/migrations/000001_create_initial_schema.up.sql
```diff
@@ -390,16 +385,27 @@ CREATE TABLE "input"
     "block_number" uint64 NOT NULL,
     "raw_data" BYTEA NOT NULL,
     "status" "InputCompletionStatus" NOT NULL,
+    "exception_data" BYTEA,
```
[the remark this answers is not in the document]
vfusco: We will keep a consolidated file for now.

internal/manager/instance.go
```diff
@@ -401,14 +413,31 @@ func (m *MachineInstanceImpl) Inspect(ctx context.Context, query []byte) (*Inspe
...
+	inspectResponse, inspectErr := fork.Inspect(inspectCtx, query)
 
 	// Create the result
 	result := &InspectResult{
 		ProcessedInputs: processedInputs,
-		Accepted:        accepted,
-		Reports:         reports,
-		Error:           inspectErr,
+	}
+	if inspectResponse != nil {
```
mpolitzer, line 418: This `if/then/else` blocks could be simplified a bit knowing that `inspectResponse` and `inspectErr` are mutually exclusive.
Something along these lines:

**UNTESTED**
```
if inspectResponse == nil && inspectErr == nil {
	return nil, errors.New("invariant violation: !(inspectResponse == nil && inspectErr == nil)")
} else if inspectErr != nil {
	result.Error = inspectErr
} else if !inspectResponse.Status.IsCompleted() {
	result.Error = errors.Join(ErrIncompleteInspect, machine.ErrMachineInternal)
} else {
    result.Reports = inspectResponse.Reports
	switch inspectResponse.Status {
	case machine.CompletionStatusAccepted:
		result.Accepted = true
	case machine.CompletionStatusRejected:
	case machine.CompletionStatusException:
		result.Error = machine.ErrException
	case machine.CompletionStatusHalted:
		result.Error = machine.ErrHalted
	default:
		result.Error = errors.Join(ErrIncompleteInspect, machine.ErrMachineInternal)
    }
}
```

internal/model/models.go
```diff
@@ -986,6 +986,7 @@ type Input struct {
 	BlockNumber        uint64                `json:"block_number"`
 	RawData            []byte                `json:"raw_data"`
 	Status             InputCompletionStatus `json:"status"`
+	ExceptionData      []byte                `json:"-"`
```
mpolitzer, line 989: No `json:"exception_data,omitempty"` instead?
(maybe not here, but on the Marshal/Unmarshal instead)

internal/repository/postgres/replay.go
```diff
@@ -0,0 +1,540 @@
...
+	stmt := table.Input.SELECT(table.Input.Index).
+		WHERE(whereInputApp.AND(replayCompletedStatus(table.Input.Status))).
+		ORDER_BY(table.Input.Index.DESC()).
+		LIMIT(1)
+	sqlStr, args := stmt.Sql()
+	var lastInputIndex uint64
+	if err := tx.QueryRow(ctx, sqlStr, args...).Scan(&lastInputIndex); err != nil {
+		return err
+	}
+	expectedLastInputIndex := total - 1
```
mpolitzer, line 195: check if the whole sequence is valid as well?
as in no skips and no repetitions.

internal/repository/replay.go
```diff
@@ -0,0 +1,208 @@
...
+		verification ReplayVerificationLevel,
+	) (model.ReplaySummary, error)
+	ReplayPage(ctx context.Context, request ReplayPageRequest) ([]*model.ReplayRecord, error)
+}
+
+// ReplayStructureViolationKind identifies one payload-free replay invariant.
+type ReplayStructureViolationKind uint8
+
+const (
+	ReplayStructureCompletedInputSequence ReplayStructureViolationKind = iota + 1
```
mpolitzer, line 92: Is there a reason for `iota + 1` instead of plain `iota`? looks a bit magic.

internal/manager/manager.go
```diff
@@ -105,124 +307,223 @@ func NewMachineManager(
...
+		}
+	}
+	return active
+}
+
+// RecordApplicationFailure immediately fences app and queues the exact
+// normalized FAILED reason for a later durability retry. Callers use this only
+// after their initial status write failed; this method does not write the
+// repository itself.
+func (m *MachineManager) RecordApplicationFailure(app *Application, reason string) {
```
mpolitzer: `InsertIntoPendingApplicationFailures` or `UpdatePendingApplicationFailures` would be better names, Record gives the idea of saving it somewhere, not the case here.

mpolitzer (dismissed):

internal/manager/instance.go
vfusco: They are intentionally not mutually exclusive. The machine Inspect contract permits a non-nil response together with an error so reports emitted before incomplete execution are preserved. The manager copies that partial response first and then gives the execution error precedence. The proposed structure would discard partial reports and its early return would skip closing the fork. This behavior is covered by ExecutionErrorPreservesPartialReports, so I suggest keeping the current structure.

internal/model/models.go
vfusco: The "-" on the struct field is required. Input has a custom MarshalJSON and the field is re-declared in the aux struct, so a tag here would emit a duplicate key. On the aux struct I kept it non-omitempty on purpose: `jsonrpc-discover.json` declares exception_data as `oneOf[ByteArray, null]` exactly like machine_hash/outputs_hash, and those are always emitted too. Clients can then rely on the key existing and just test for null rather than distinguishing "absent" from "null".

internal/repository/postgres/replay.go
vfusco: Yes. The check uses the completed-row count together with the maximum completed index. Because input indexes are non-negative and unique by primary key, N completed rows with maximum index N-1 can only be the complete sequence 0..N-1; a skip would force the maximum higher, and repetitions cannot be stored. I agree this proof is not obvious from the code, so I’ll add a comment and an explicit gap test.

internal/repository/replay.go
vfusco: Zero is intentionally reserved so the zero value cannot identify a real violation; an uninitialized or unsupported kind therefore formats as unknown. I agree iota + 1 does not explain that by itself, so I’ll document the reserved zero value.

internal/manager/manager.go
vfusco: I agree that Record can sound like durable storage. I would avoid InsertIntoPendingApplicationFailures or UpdatePendingApplicationFailures, since those expose the internal collection and the method does not update existing entries. The domain operation is installing a local safety fence while persistence is retried, so FenceApplicationFailure would be clearer. I'll rename it accordingly.

renatomaia (approved):

mpolitzer (approved):

PR #795, opened 2026-09-02
Preserve terminal machine outcomes and prepare complete state proofs

vfusco:
This PR prepares the node for the next contracts update and improves terminal machine handling.

- Generate and store complete proofs for the TX buffer, `iflags.Y`, and HTIF `tohost`.
- Preserve the post-run machine state for terminal outcomes.
- Add `GUEST_EXCEPTION`, `MACHINE_HALTED`, `MCYCLE_OVERFLOW`, and `UNEXPECTED_YIELD` application states.
- Stop input execution after a terminal outcome, including after a restart.
- Keep L1 event indexing active for terminal applications.
- Expose the new states and proof data through JSON-RPC.
- Add unit and integration tests for terminal outcomes and restart behavior.

This PR does not submit the new proof fields to the current contracts.


internal/repository/postgres/schema/migrations/000001_create_initial_schema.up.sql
```diff
@@ -215,8 +264,12 @@ CREATE TABLE "epoch"
...
     "input_index_upper_bound" uint64 NOT NULL,
     "machine_hash" hash,
-    "outputs_merkle_root" hash,
-    "outputs_merkle_proof" BYTEA[],
+    "tx_buffer_data_block" hash,
+    "tx_buffer_proof" BYTEA[],
+    "iflags_y_data_block" hash,
+    "iflags_y_proof" BYTEA[],
+    "htif_tohost_data_block" hash,
+    "htif_tohost_proof" BYTEA[],
```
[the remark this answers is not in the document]
vfusco: This is still alpha software. we will keep one migration for now cause it easier to maintain. There is currently no demand for migration. And note that the node is tied to a machine emulator version and there's no compatibility or migration between these versions either.

pkg/machine/implementation.go
```diff
@@ -88,8 +90,26 @@ const (
...
 const maxReports = 65536 // 2^16
 
-const TxBufferAddress uint64 = 0x60800000
-const HashLog2Size = 5 // 32 bytes
+const (
+	// These addresses and hash-tree sizes are defined by the Cartesi Machine
+	// emulator.
+	iflagsYAddress          uint64 = 0x308
+	htifTohostAddress       uint64 = 0x330
+	TxBufferAddress         uint64 = 0x60800000
```
renatomaia, line 98: Can we define these values from the constants provided from the Cartesi Machine? For instance `TxBufferAddress uint64 = emulator.CmioTxBufferStart`?

pkg/machine/implementation.go
```diff
@@ -141,51 +161,154 @@ func (m *machineImpl) Hash(ctx context.Context) (Hash, error) {
...
+	return nil
 }
 
-func (m *machineImpl) OutputsHashProof(ctx context.Context) ([]Hash, error) {
+func (m *machineImpl) readLeafProof(ctx context.Context, machineHash Hash, wordAddress uint64) (LeafProof, error) {
 	if err := checkContext(ctx); err != nil {
-		return nil, err
+		return LeafProof{}, err
+	}
+	dataBlockAddress := wordAddress &^ ((uint64(1) << HashLog2Size) - 1)
```
renatomaia, line 219: Should the tests use this expression with new const `HashLog2Size` instead of hardcoded `31`? See an example [here](https://github.com/cartesi/rollups-node/blob/fad13fb01c78e529c284ce6de615116aaa4fc91f/pkg/machine/implementation_test.go#L228).

pkg/emulator/types.go
```diff
@@ -4,6 +4,8 @@
 // This package is a binding to the emulator's C API.
 // Refer to the machine-c files in the emulator's repository for documentation
 // (mainly machine-c-api.h and jsonrpc-machine-c-api.h).
+//
+//nolint:revive // Exported ALL_CAPS names intentionally mirror the emulator C API.
```
mpolitzer: Explanation doesn't match what is actually being done.
Exported names are camel case, not ALL_CAPS / SCREAMING_SNAKE_CASE / UPPER_SNAKE_CASE.

pkg/machine/backend.go
```diff
@@ -29,6 +29,18 @@ type HashCollectorState struct {
 	ConsoleIOError        string
 }
 
+// MemoryProof is the complete Merkle proof returned by the emulator for a
+// memory range. Keeping the root and target metadata allows callers to verify
+// that a proof was produced for the requested machine state and address.
+type MemoryProof struct {
```
mpolitzer, line 36: Should we follow the memory range terminology here and use `MemoryRangeProof` instead?

pkg/machine/implementation.go
vfusco: These constants stay local to keep the implementation independent of the libcartesi backend. `TestStateProofConstantsMatchEmulator` checks all values against the linked emulator. It also uses `GetRegAddress` to check the two register addresses.

pkg/machine/implementation.go
vfusco: Yes. This value duplicates the data-block size. I will replace it with an expression based on `HashLog2Size`. I will also replace the hardcoded `32` in this helper.

internal/advancer/advancer.go
```diff
@@ -323,18 +362,40 @@ func (s *Service) processInputs(ctx context.Context, app *Application, inputs []
...
+			)
+			return processed, true, nil
 		}
 
 		// Create a snapshot if needed
 		if result.Status == InputCompletionStatus_Accepted {
 			err = s.handleSnapshot(ctx, app, machine, input)
 			if err != nil {
-				if errors.Is(err, context.Canceled) {
+				switch {
```
mpolitzer: instead of a `if err != nil { ... }`, the switch could have a `case nil` at the start to simplify this a bit.

Minor point

internal/model/models.go
```diff
@@ -397,6 +388,26 @@ func (e ApplicationStatus) IsTerminal() bool {
...
+		ApplicationStatus_MachineHalted,
+		ApplicationStatus_McycleOverflow,
+		ApplicationStatus_UnexpectedYield:
+		return true
+	case ApplicationStatus_OK,
+		ApplicationStatus_Failed,
+		ApplicationStatus_Diverged,
+		ApplicationStatus_Corrupted:
+		return false
+	default:
```
mpolitzer: I would remove the `default:` branch and move the `return false` to after the switch. This way the compiler gives a warning on new `ApplicationStatus` entries.

pkg/emulator/types.go
vfusco: The new names use CamelCase. The suppression is for the existing CM_JSONRPC_* and REG_* names. I will move each directive to the applicable constant block and update the comment.

internal/advancer/advancer.go
vfusco: Agreed. I will add case err == nil and remove the outer if. This makes the control flow flatter.

internal/model/models.go
vfusco: Agreed. I will remove the default branches from both status switches. The exhaustive linter will then report new enum values.

pkg/machine/backend.go
vfusco: The emulator exposes this value as `Proof`, and its underlying type is `hash_tree_proof`. The generic name works there because `machine.get_proof` provides the required context. In `pkg/machine`, an exported `Proof` type would be ambiguous because the package also contains `StateProof` and `LeafProof`. `MemoryProof` identifies what the proof authenticates: an aligned span of machine memory. `MemoryRangeProof` could imply an emulator-configured memory range, which is a different concept. `HashTreeProof` describes the mechanism but not the authenticated data. For these reasons, I prefer to keep `MemoryProof` and improve its comment.

mpolitzer (approved):

